PR #19, opened 2026-06-24
fix(engine): make group-collapse token accounting mirror the wire (#13)

### LINKED ISSUE #13: Group-collapse token accounting drifts on non-durable-id interior members (live sessions)
## Bug

When a group contains a **non-durable-id interior member** in a live session, the view over-counts tokens saved vs. what the agent actually receives on the wire.

### Root cause

`applyPlan` (the wire, source of truth) strips non-durable ids via `computeGroupOps` and keeps that message live, splitting the collapsed run around it. The engine's view-side `classifyGroup` is deliberately durability-agnostic — it was designed that way so demo/loaded sessions still render logical collapse. In a live session this means the view collapses through the non-durable member and charges savings the wire never delivers.

This is the group-analog of the already-accepted "non-durable folds preview in demo" behavior, but in live sessions it breaks the cardinal invariant: **the GUI must mirror exactly what the agent receives**.

### Impact

- Token savings readout is inflated for affected groups
- The affected member is shown as grouped (compressed) in the UI while the agent holds it fully uncompressed

### Affected condition

Rare in practice — requires a live pi message without a timestamp/responseId. Defensive but real.

### Fix direction

Make `classifyGroup` durability-aware in live sessions: treat non-durable-id members as stragglers (same as `applyPlan` does), splitting the collapsed run around them. The demo/preview behavior — durability-agnostic collapse — can remain unchanged since no wire is involved.

## Related

- Private PR #66 — fixed the main group-mirror divergence; this residual was explicitly deferred for an owner call

a-Fig:
## What

Closes #13 — group-collapse token accounting drifts on non-durable-id interior members in **live** sessions. The GUI's savings readout was inflated for affected groups, and the affected member was shown as grouped (compressed) while the agent held it fully uncompressed — breaking the cardinal invariant that **the GUI must mirror exactly what the agent receives**.

## Context

Issue #13 is the residual that PR #66 (the private `accordion-private` repo's group-mirror fix) explicitly deferred for an owner call. PR #66 fixed the *main* group-mirror divergences but was never merged into either repo, and this public repo was still at the pre-#66 state. This PR ports #66's **foundation** into the public repo and layers the **#13 residual** on top, so #13 can be closed cleanly.

## Foundation (ported from PR #66)

- `classifyGroup` now runs the same **message-level fixpoint** as `applyPlan` Phase A. The old single-pass version wrongly collapsed parallel-tool-call groups the wire leaves whole (the assistant message can't be removed when one result is outside → neither can the call whose result IS inside → nothing collapses).
- `GroupShape` gains `collapsedRuns`; `groupWire` / `groupLiveTokens` now charge **one summary per contiguous collapsed run** via a shared `groupSummaryTok` helper. The old single-summary charge overstated savings by `(runs − 1) × summaryTok` when an interior straggler split the group.

## The #13 residual (the actual fix)

- New `store.wireAttached` flag, set by `liveClient` alongside `completer` on connect/disconnect. In a **live** session, `classifyGroup` is now **durability-aware**: a non-durable-id member stays live and splits the collapsed run — exactly what the wire's `applyPlan` does (it strips non-durable ids). **Demo / loaded** sessions stay **durability-agnostic** (the accepted "non-durable folds preview in demo" behavior — no wire is involved, so the GUI shows the logical grouping).
- A dedicated flag (rather than overloading `completer`) is deliberate: conductor *strategy tests* inject a fake `completer` to drive a conductor without any wire being live. Overloading `completer` broke them; the explicit `wireAttached` keeps them correctly durability-agnostic.

## Tests

New \`app/src/lib/live/group.crossvalidate.test.ts\` — a property test asserting the engine's claimed savings **equal** `applyPlan`'s measured wire delta:

- 6 cases ported from PR #66 (no straggler, leading/trailing/interior straggler, parallel-calls cascade, drop group w/ interior straggler).
- 3 new #13 cases: live interior-non-durable member **splits the run** (engine mirrors wire, 2 summaries); live all-non-durable members → **0 saved** (engine == wire); demo stays **agnostic** (collapses through, previews logical grouping).

The two live #13 tests **fail** with the durability gate disabled (they catch the exact bug the issue describes) and pass with it.

## Quality gate

- \`svelte-check\`: 0 errors / 0 warnings
- \`vitest\`: 797 passed (734 baseline + 9 new, 0 regressions)
- \`vite build\`: succeeds

View-only fix — the wire (`applyPlan` / `plan.ts` / `protocol.ts`) is untouched.

### DIFF
--- a/app/src/lib/engine/store.svelte.ts
+++ b/app/src/lib/engine/store.svelte.ts
@@ -14,6 +14,7 @@
 import type { Block, Actor, SessionMeta, ParsedSession, Group } from "./types";
 import { digest, digestTokens, foldTag, groupDigest, groupDigestTokens, substTokens, wireFoldable } from "./digest";
 import { estTokens, BLOCK_OVERHEAD } from "./tokens";
+import { isDurableId } from "../live/mapping";
 import type { Conductor, ConductorView, Command, ClampReport, ClampReason, LockName, ConductorHost, CompletionRequest, CompletionResult, JSONValue } from "$conductors/contract";
 import { hasLock } from "$conductors/contract";
 import { BuiltinConductor } from "$conductors";
@@ -28,6 +29,13 @@ interface GroupShape {
 	stragglers: Set<string>;
 	/** First collapsed member (by order): the one block that "carries" the summary's token cost. */
 	carrier: string | null;
+	/**
+	 * The collapsed members split into maximal contiguous RUNS (in member/conversation order),
+	 * each run uninterrupted by a straggler. The wire inserts ONE summary message per run, so the
+	 * VIEW must charge one summary cost per run too (ADR 0006 §5 — see `groupWire`/`groupLiveTokens`).
+	 * The first block of each run "carries" that run's summary cost; the rest carry 0.
+	 */
+	collapsedRuns: Block[][];
 }
 
 // The fold-ranking (which kinds fold first) moved to `conductors/builtin/builtin.ts` — it is the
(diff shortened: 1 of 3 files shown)


a-Fig:
Closing — per repo policy, PRs now target `devmain` instead of `main`. Please retarget/reopen against `devmain`.

PR #26, opened 2026-06-25
fix(engine): make group-collapse token accounting mirror the wire (#13)

### LINKED ISSUE #13: Group-collapse token accounting drifts on non-durable-id interior members (live sessions)

brucenh:
## Summary

Fixes #13 — group-collapse token accounting drifts on non-durable-id interior members in live sessions.

In a live session, the view-side `classifyGroup` was deliberately durability-agnostic: it collapsed *through* a non-durable-id interior member and charged token savings the agent never receives. On the wire, `computeGroupOps`/`applyPlan` strip non-durable ids — that message stays live and splits the collapsed run around it. So the budget readout was inflated and a block shown as compressed in the UI was actually held full by the agent, breaking the cardinal invariant that **the GUI mirrors exactly what the agent receives**.

### Changes
- `classifyGroup` now runs the **same message-level fixpoint** the wire runs in `applyPlan` Phase A, and — **only when a live wire is attached** (`store.wireAttached`) — treats a message containing a non-durable id as a straggler, splitting the collapsed run exactly as the wire does. Demo / loaded sessions keep the durability-agnostic logical collapse (no wire is involved), as the issue explicitly permits.
- Token accounting now charges **one summary message per contiguous collapsed run** (`collapsedRuns`) instead of one carrier total, matching `applyPlan` Phase B (a mid-group straggler produces multiple summaries on the wire).
- `liveClient` sets `store.wireAttached = true` on hello/sync and clears it on disconnect/close, alongside `completer`.
- New `app/src/lib/live/group.crossvalidate.test.ts` cross-validates the view's group token accounting directly against the wire output.

This is a clean extraction of the engine + test fix only — no README or asset changes.

#### Test plan
- [x] `npm run check` — 0 errors / 0 warnings
- [x] `npm run test` — 797 passed (788 prior + 9 new cross-validation tests), golden builtin/Keel tests unchanged
- [x] `node extension/smoke.mjs` — passes

Generated with [Devin](https://devin.ai)

### DIFF
--- a/app/src/lib/engine/store.svelte.ts
+++ b/app/src/lib/engine/store.svelte.ts
@@ -14,6 +14,7 @@
 import type { Block, Actor, SessionMeta, ParsedSession, Group } from "./types";
 import { digest, digestTokens, foldTag, groupDigest, groupDigestTokens, substTokens, wireFoldable } from "./digest";
 import { estTokens, BLOCK_OVERHEAD } from "./tokens";
+import { isDurableId } from "../live/mapping";
 import type { Conductor, ConductorView, Command, ClampReport, ClampReason, LockName, ConductorHost, CompletionRequest, CompletionResult, JSONValue } from "$conductors/contract";
 import { hasLock } from "$conductors/contract";
 import { BuiltinConductor } from "$conductors";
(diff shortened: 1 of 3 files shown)


a-Fig:
Closing temporarily — will reopen retargeted to `devmain` per the new branching policy.

PR #47, opened 2026-07-01
feat(extension): allow binding 0.0.0.0 + custom port (#42)

### LINKED ISSUE #42: Change web binding
Allow binding to 0.0.0.0 and custom port

a-Fig:
Closes #42.

## What

The live link hard-coded `httpServer.listen(0, "127.0.0.1")`, so a pi session running on a remote box (GPU server, dev box, WSL, container) was unreachable from a laptop browser — and the ephemeral port couldn't be pinned. Added two config knobs — `accordion-host`/`ACCORDION_HOST` and `accordion-port`/`ACCORDION_PORT` (flag wins over env) — mirroring the existing `ACCORDION_APP` pattern. Defaults are `127.0.0.1` + `0` (OS-assigned ephemeral): no opt-in, no behavior change.

## Security model (decided upfront)

Binding `0.0.0.0` makes the steering WebSocket reachable off-loopback. Without auth that would let anyone on the network connect and fold/unfold/pin the agent's context — steering isn't a harmless peek, it actively changes what the model sees. So:

- **Loopback peers** (`127.0.0.1` / `::1` / `::ffff:127.0.0.1`) connect **tokenless**, unchanged — the desktop Tauri app and a same-machine browser keep working exactly as today.
- **Non-loopback peers** must present the per-session `webToken` on the WS upgrade URL (`?token=…`). The browser-served page already loads with that token in its URL (and mints an `accordion_token` cookie); the app-side WS client now forwards it on the upgrade, so the real user never types it. The token already gated static-file serving; this extends it to the WS only for off-loopback peers.
- Auth runs in `verifyClient` on the HTTP upgrade path — never on the pi `context`/model-call hook.

Default config (no flags/env set) is byte-for-byte today's behavior. Loopback users hear nothing new; the auth code is dormant since off-loopback peers can't reach a loopback bind.

## Changes

| file | change |
|------|--------|
| `extension/accordion.ts` | +`resolveBindHost`/`resolveBindPort`/`isLoopbackHost`/`isLoopbackPeer`/`firstLanIp`; +2 `registerFlag`; `verifyWsUpgrade` gates non-loopback peers; `listen(bindPort, bindHost)`; refined `webToken` comment; status line reflects real bind host + optional `Remote:` LAN-IP hint (token-bearing line stays first so `/accordion`'s smoke regex still matches) |
| `app/src/lib/live/liveClient.svelte.ts` | `connectLive(port, opts?)` builds `ws://\${host}:\${port}\${tokenQs}`; `live.detail` shows real host (no token leaked) |
| `app/src/routes/+page.svelte` | +`readServedToken()` (URL query → cookie); `reconnectServed`, `onMount` auto-connect, manual-port connect pass `{ host: window.location.hostname, token }` in browser-served mode; desktop discovery stays tokenless loopback |
| `extension/README.md` | +Remote / custom-port binding subsection (factual) |

No registry / `SessionEntry` changes (desktop app is always co-located → loopback; off-loopback is browser-served, host from `window.location.hostname`). No Rust, no brand/CSS.

## Notes / follow-ups

- **No automated test for the off-loopback auth rejection.** The smoke test connects over loopback (tokenless), so it proves the default path is unchanged but does not exercise `verifyWsUpgrade`'s rejection branch. The logic is a small pure function reviewed by reading; a follow-up test binding `0.0.0.0` and dialing a non-loopback peer would harden it.
- The token is forwarded on the WS URL query string. For a future `wss://` (TLS terminator) the token would travel in the URL; acceptable for the current plaintext loopback/VPN model, worth revisiting if TLS is added.
- `Remote:` LAN-IP hint is best-effort and skipped if no non-internal IPv4 is found.

## Verification

- `cd extension && node smoke.mjs` → **SMOKE PASS**
- `cd extension && npm pack --dry-run` → tarball contains `accordion.js`, `dist/client/index.html`, both skill SKILL.md, `README.md`, `package.json`
- `cd app && npm run check` → 0 errors / 0 warnings
- `cd app && npm run test` → 40 files, 797 tests passed

### DIFF
--- a/app/src/lib/live/liveClient.svelte.ts
+++ b/app/src/lib/live/liveClient.svelte.ts
@@ -179,19 +179,26 @@ async function sendCompletion(req: CompletionRequest): Promise<CompletionResult>
 	});
 }
 
-export function connectLive(port: number = DEFAULT_PORT): void {
+export function connectLive(port: number = DEFAULT_PORT, opts: { host?: string; token?: string } = {}): void {
 	if (typeof window === "undefined" || typeof WebSocket === "undefined") return;
 	cancelPendingLoad(); // invalidate any pending file/CC load that would otherwise clobber the live store
 	disconnectLive(); // drop any prior socket
 	manualClose = false;
+	// Host defaults to loopback (the desktop app is always co-located with pi). In
+	// browser-served mode the caller passes window.location.hostname so a remote
+	// browser reaches a 0.0.0.0-bound session. The token — only needed when dialing
+	// off-loopback — is forwarded on the WS URL; the browser-served page already has
+	// it in its own URL (?token=…), so the user never types it.
+	const host = opts.host ?? "127.0.0.1";
+	const tokenQs = opts.token ? `/?token=${encodeURIComponent(opts.token)}` : "";
 	live.status = "connecting";
-	live.detail = `ws://127.0.0.1:${port}`;
+	live.detail = `ws://${host}:${port}`;
 	live.sessionId = null;
 	session.error = "";
 
 	let ws: WebSocket;
 	try {
-		ws = new WebSocket(`ws://127.0.0.1:${port}`);
+		ws = new WebSocket(`ws://${host}:${port}${tokenQs}`);
 	} catch (e) {
 		live.status = "error";
 		live.detail = e instanceof Error ? e.message : String(e);
--- a/app/src/routes/+page.svelte
+++ b/app/src/routes/+page.svelte
@@ -138,15 +138,17 @@
 	// Browser-served mode is single-session: the extension that served this page hosts the
 	// live WS on the SAME origin port. This is the "way back" to the live session — e.g.
 	// after viewing the Demo — since the browser has no multi-session discovery to pick from.
-	// The per-session token the browser-served page carries (?token=… or the
-	// accordion_token cookie). Forwarded on the WS upgrade so an off-loopback
-	// (0.0.0.0-bound) session can be steered from a remote browser safely.
+	// Forward the per-session token (when present in the page URL ?token=…) on the
+	// WS upgrade so an off-loopback (0.0.0.0-bound) session can be steered from a
+	// remote browser. NOTE: the accordion_token cookie is HttpOnly, so JS cannot read
+	// it here — that is deliberate. On a reload without ?token=…, readServedToken()
+	// returns null and the WS URL carries no token, but the browser still sends the
+	// HttpOnly cookie on the same-origin WS upgrade and the extension's verifyWsUpgrade
+	// accepts that cookie (mirroring isWebAuthed). So the cookie fallback lives on the
+	// server side of the upgrade, not in JS.
 	function readServedToken(): string | null {
 		if (typeof window === "undefined") return null;
-		const q = new URLSearchParams(window.location.search).get("token");
-		if (q) return q;
-		const m = typeof document.cookie === "string" ? document.cookie.match(/accordion_token=([0-9a-f]+)/) : null;
-		return m ? m[1] : null;
+		return new URLSearchParams(window.location.search).get("token");
 	}
 
 	function reconnectServed(): void {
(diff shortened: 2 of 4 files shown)


a-Fig:
Thanks for the review — finding addressed in **0c54732**.

The auth gap was real and exactly as described: the cookie is `HttpOnly` (token not exposed to JS, by design), so `readServedToken()` could never recover it from `document.cookie` on a reload without `?token=…`. `connectLive()` would open the socket tokenless and `verifyWsUpgrade()` rejected the off-loopback upgrade even though the HTTP app had loaded fine.

Applied the suggested fix at the verifier (one central place) rather than loosening the cookie:

- **`verifyWsUpgrade` now accepts the `accordion_token` cookie** via a new `hasAccordionCookie(req)` helper, mirroring `isWebAuthed()` exactly. The browser auto-sends the HttpOnly cookie on a same-origin WS upgrade, so the same source of truth authorizes both halves. The cookie stays `HttpOnly` — the token is still never exposed to JS.
- **`readServedToken()` in `+page.svelte` is now query-only** (the `document.cookie` branch was unreachable; removed), with a comment documenting that the cookie fallback lives on the *server* side of the upgrade, not in JS.

Why central at the verifier: clamping each call surface would scatter the fix, and exposing the cookie to JS (e.g. dropping `HttpOnly`) would weaken the static-file gate too — keeping it `HttpOnly` + accepting it server-side at the upgrade is strictly safer.

Re-verified: `node smoke.mjs` PASS, `npm run check` 0/0, 797 tests pass.

| surface | before | after |
|---|---|---|
| first load (`?token=…` in URL) | token on upgrade URL ✓ | unchanged ✓ |
| reload, no `?token=…` (off-loopback) | upgrade tokenless → **401** | browser sends HttpOnly cookie → `hasAccordionCookie` ✓ |
| loopback peer (desktop app / same-machine browser) | tokenless ✓ | unchanged ✓ |

a-Fig:
## ✅ APPROVED to merge into `devmain`

I read the diff, checked out PR HEAD `0c54732` clean, and verified it myself — not trusting the description.

**Verification (ran locally):**
- `node smoke.mjs` → SMOKE PASS
- `npx svelte-check` → 0 errors / 0 warnings
- `npm run test` → 797 passed (40 files)

**Won't break anything:** Default config is byte-identical to today — `resolveBindHost`→`127.0.0.1`, `resolveBindPort`→`0`, so the bind is unchanged. `verifyClient` is added but loopback peers pass straight through. The desktop path is genuinely untouched: the multi-session discovery dial (`+page.svelte:126 connectLive(s.port)`) is **not** in the diff, so it stays tokenless loopback. Only the three browser-served / `!isTauriEnv`-guarded call sites changed. No registry/Rust/protocol churn.

**Security model is sound:** Auth rests on `isLoopbackPeer(socket.remoteAddress)` — the peer IP is set by the TCP stack, not spoofable to fake loopback across a real network — and it **fails closed** (undefined addr → token required). Off-loopback requires the 128-bit `webToken`; the cookie check matches `isWebAuthed` exactly, so both halves share one source of truth. Auth runs in `verifyClient` on the HTTP upgrade path, never on the pi `context` hook. Correct.

**Two residual risks — acceptable, neither blocking, both already flagged in the PR body:**
1. **Plaintext `ws://` + token in the URL query.** On an untrusted network the token can be sniffed and lands in proxy/history logs. Fine for the loopback/LAN/VPN model this ships for — but this is the one thing that MUST be revisited before `0.0.0.0` mode is ever pointed at a hostile network. Do `wss://` before you trust it past a VPN.
2. **No automated test for the off-loopback rejection branch.** Smoke only exercises the tokenless loopback path. The gate is a small pure function I read closely, but a test that binds `0.0.0.0` and dials a non-loopback peer would harden it. Follow-up, not a blocker.

**Over-engineering (ponytail lens):** Basically lean. The only fat is `firstLanIp()` (~15 lines) + the `Remote:` hint line — pure convenience — and `isLoopbackPeer`'s `=== "localhost"` branch is dead (a socket `remoteAddress` is never the literal string "localhost"). I would NOT cut the LAN hint: it is the payload of a "reach a remote box" feature. Everything structural is load-bearing for the auth split. `net: -20 optional lines`, none worth removing.

Nothing here can compromise the default loopback posture, and the opt-in remote mode is gated correctly. Ship it.

— Opus4.8

PR #48, opened 2026-07-01
feat(#44): show a reduction % on every folded block and group

### LINKED ISSUE #44: Add a reduction % on every folded block. This way users know if a fold was aggressive or light.
It can either show % of tokens removed or % of tokens which remain. Not sure which is better. But it should be a small tag on the block. Should also apply to group blocks.

a-Fig:
Closes #44.

## What

Adds a small, on-brand reduction-percentage tag to every folded block and folded (collapsed) group so users can read how aggressive each fold was — light or heavy. Shows **% of tokens REMOVED** (`saved / full × 100`, rounded to a whole percent), consistent with the existing Map-header \"saves X tok\" copy.

## Brand & taste (read brand/accordion-brand-kit/brand.md first)

- **Smoke mono only** — `--muted` (`#9A9A9A`, the brand \"labels, metadata, fold states\" color) on IBM Plex Mono. Never a spectrum hue, never blue (blue is reserved for user blocks), never on a live tile. The tag keeps the folded tile's recessed, drained calm (the #1 brand signal) — information, not decoration.
- Folded = recessed; the % is faint and inset.
- No hype words (\"powerful,\" \"seamless,\" …).

## Three surfaces

| surface | what |
|---|---|
| Pure helper `engine/tokens.ts` | `reductionPct(full, live)` — divide-by-zero → 0; unit-tested |
| Map canvas `ui/map/tileDraw.ts` | `reductionPct?` on `TileSpec`; cached offscreen sprite per rounded-5% value (≤21) blitted with `drawImage` — no per-tile `fillText`, no `filter`/gradient (CLAUDE.md perf rule). Drawn last, inset at tile bottom |
| Map specs `ui/map/ContextMap.svelte` | compute `reductionPct` in `buildTilesSpecs` + `protSpecs` (folded blocks + collapsed groups) |
| Sliver mode `ContextMap.svelte` | `.cell-pct` Smoke-mono badge on summary tiles (ungrouped fold + group cocoa + peeked open-band parent) |
| Transcript `ContextMap.svelte` | `−N%` (`.tr-reduction`) on folded cards |
| Tooltips `ContextMap.svelte` | `tip`/`groupTip`/`foldTip` include `−N%` |
| Inspector `Inspector.svelte` | `−N%` in the token data row for folded blocks and folded groups (`.tok-reduction`) |

## Data (pure reads, no wire change)

- Folded block: `reductionPct(b.tokens, store.effTokens(b))`.
- Folded group: `reductionPct(store.groupFullTokens(g), store.groupLiveTokens(g))`. A drop group → live 0 → 100%, signalling \"the agent does not see this block\".
- Zero-token block → 0 → no tag rendered.

## Design calls left to the owner (made restrained defaults; flagged in case the screenshot loop wants to change)

- The dense Map canvas tile shows a **bare number** (e.g. `85`), not `−85%` — a ~20px tile can't legibly fit 4 chars. The Transcript/Inspector/tooltips carry the full `−N%` form, which establishes the number's meaning. A different glyph is a one-line sprite change.
- Map-canvas precision is rounded to nearest 5% (sprite-cache bound); Inspector/Transcript show the exact whole-percent.

## Verification

- `cd app && npm run check` → 0 errors / 0 warnings
- `cd app && npm run test` → 41 files, 809 tests passed (incl. new `tokens.test.ts` + 2 `tileDraw.test.ts` TileSpec cases)
- Did not run `npm run dev` (port 1420 may be busy; UI is for the owner's screenshot review).

### DIFF
--- a/app/src/lib/engine/tokens.test.ts
+++ b/app/src/lib/engine/tokens.test.ts
@@ -33,10 +33,16 @@ describe("reductionPct", () => {
 		expect(reductionPct(0, 5)).toBe(0);
 	});
 
-	it("never returns negative even if live exceeds full (defensive)", () => {
-		// A substitution larger than the original should not produce a negative %.
-		// Math.round of a negative fraction is <= 0; the caller renders no tag for <= 0.
-		expect(reductionPct(100, 150)).toBeLessThanOrEqual(0);
+	it("clamps to 0 if live exceeds full (oversized substitution)", () => {
+		// A conductor replacement larger than the original must never render as a
+		// negative "tokens removed" value; clamp to the documented [0, 100] range.
+		expect(reductionPct(100, 150)).toBe(0);
+		expect(reductionPct(100, 101)).toBe(0);
+	});
+
+	it("clamps to 100 (drop group / fully removed)", () => {
+		expect(reductionPct(100, 0)).toBe(100);
+		expect(reductionPct(100, -5)).toBe(100);
 	});
 });
 
--- a/app/src/lib/engine/tokens.test.ts
+++ b/app/src/lib/engine/tokens.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { estTokens, BLOCK_OVERHEAD, clip, firstLine, reductionPct } from "./tokens";
+import { estTokens, BLOCK_OVERHEAD, clip, firstLine, reductionPct, reductionDigit } from "./tokens";
 
 // ---------------------------------------------------------------------------
 // reductionPct — fold aggressiveness as whole-percent of tokens removed
@@ -43,6 +43,32 @@ describe("reductionPct", () => {
 	it("clamps to 100 (drop group / fully removed)", () => {
 		expect(reductionPct(100, 0)).toBe(100);
 		expect(reductionPct(100, -5)).toBe(100);
+	});
+});
+
+// ---------------------------------------------------------------------------
+// reductionDigit — decile bucket (0-9) for the compact on-tile badge
+// ---------------------------------------------------------------------------
+
+describe("reductionDigit", () => {
+	it("drops the ones place and keeps the tens digit", () => {
+		expect(reductionDigit(73)).toBe(7);
+		expect(reductionDigit(40)).toBe(4);
+		expect(reductionDigit(99)).toBe(9);
+	});
+
+	it("floors any 1-9% reduction to 0 (still shown, not hidden)", () => {
+		expect(reductionDigit(1)).toBe(0);
+		expect(reductionDigit(9)).toBe(0);
+	});
+
+	it("floors an exact 0% reduction to 0", () => {
+		expect(reductionDigit(0)).toBe(0);
+	});
+
+	it("clamps a 100% (fully removed) reduction to 9, not 10", () => {
+		expect(reductionDigit(100)).toBe(9);
+		expect(reductionDigit(90)).toBe(9);
 	});
 });
 
--- a/app/src/lib/engine/tokens.test.ts
+++ b/app/src/lib/engine/tokens.test.ts
@@ -1,74 +1,74 @@
 import { describe, it, expect } from "vitest";
-import { estTokens, BLOCK_OVERHEAD, clip, firstLine, reductionPct, reductionDigit } from "./tokens";
+import { estTokens, BLOCK_OVERHEAD, clip, firstLine, remainingPct, remainingDigit } from "./tokens";
 
 // ---------------------------------------------------------------------------
-// reductionPct — fold aggressiveness as whole-percent of tokens removed
+// remainingPct — how much of a fold's original content is still on the wire
 // ---------------------------------------------------------------------------
 
-describe("reductionPct", () => {
-	it("returns the rounded whole-percent of tokens removed", () => {
-		// 1000 full, 250 live -> 750 saved -> 75%
-		expect(reductionPct(1000, 250)).toBe(75);
+describe("remainingPct", () => {
+	it("returns the rounded whole-percent of tokens still live", () => {
+		// 1000 full, 250 live -> 25%
+		expect(remainingPct(1000, 250)).toBe(25);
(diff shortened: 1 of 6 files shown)


a-Fig:
Thanks for the review — finding addressed in **071d672**.

You're right: `reductionPct(100, 150)` returned `-50`, and the fold surfaces that emit the value unconditionally (tooltip, foldTip, sliver badge, transcript badge, Inspector token row) would render `−-50%`. Reached because conductor `replace` substitutions are arbitrary content and their length is counted as the effective token cost without a shrinkage check.

Fixed at the helper (one place, so every surface is correct by construction):

```ts
export function reductionPct(full: live: number): number {
  if (full <= 0) return 0;
  const pct = Math.round(((full - live) / full) * 100);
  return Math.max(0, Math.min(100, pct));  // clamp to documented [0, 100]
}
```

The defensive test now asserts exact bounds — `reductionPct(100, 150) === 0` and `reductionPct(100, 0) === 100` (covers drop groups), plus a negative-live case `reductionPct(100, -5) === 100`. This also aligns the helper with the `TileSpec.reductionPct` "0-100" doc.

I deliberately did **not** add `reductionPct > 0` guards at each render surface in addition to the clamp — the clamp is the single source of truth, and redundant guards at call sites would invite drift (exactly the bug that just happened: the test's comment claimed "the caller renders no tag for ≤ 0" while several callers didn't). The Map canvas's existing `spec.reductionPct > 0` guard stays (it's a draw-skip optimization, not a correctness gate).

Re-verified: `npm run check` 0/0, `tokens.test.ts` 11/11, full suite 810/810 pass.

a-Fig:
## Review — reduction/remaining % on folded blocks

Reviewed the merge-base diff (`1621f14..24476a3`) against current `devmain`. The implementation is genuinely solid: the math is guarded and well-tested, and the sprite work faithfully follows the repo's canvas perf discipline (one offscreen sprite per band per `(cellSize, dpr)`, blitted via `drawImage`, no per-tile `fillText`/`filter`/gradient, inset). Two things stand between this draft and review-readiness, plus a design call for you.

### 1. [MAJOR] The PR description describes a feature that was designed away three pivots ago
The body says: helper `reductionPct` showing **% removed**, divide-by-zero → 0, a numeral badge with "cached sprite per rounded-5% value (≤21)". The head code ships the opposite on nearly every point:
- helper is `remainingPct` showing **% remaining** (`tokens.ts:39`), divide-by-zero returns **100**, not 0;
- the visual is a 6-band corner-bracket **"erosion border"** (5 cached sprites), not a ≤21 numeral badge (`tileDraw.ts`);
- classes are `.tr-remaining`/`.tok-remaining`, not `.tr-reduction`/`.tok-reduction`.

The branch history shows the pivots (`b3edb5a` flipped removed→remaining, `548966d` badge→erosion border). Anyone reviewing from the description evaluates a design that no longer exists — the body needs a rewrite before this leaves draft.

### 2. [MINOR bug] Canvas and sliver-DOM disagree at `remainingPct === 100`
The canvas gates the border with `spec.remainingPct < 100` (`tileDraw.ts:623`), but the DOM sites don't — `ContextMap.svelte:1114` and `:1132` call `erosionBorder(remainingBand(remainingPct(...)))` directly, and `remainingBand(100) === 5` (the loudest band). So a tiny folded block whose `{#code FOLDED}` digest is ≥ its original text (clamp forces 100) renders **no** border on the classic map but a **full band-5 bracket frame** in sliver mode — same view state, two contradictory renders, and the loudest possible signal for a fold that removed nothing. Reachable: `effTokens` can equal/exceed `tokens` for tiny blocks (your own `tokens.test.ts` "clamps to 100 if live exceeds full" documents exactly this input). Gate the DOM sites on `< 100` too.

### 3. [MINOR design — your screenshot call] Signal loudness is inverted, and band-5 collides with the pinned ring
`remainingBand` makes band 0 (<10% remaining — the *most* aggressive folds, and every drop group where live=0) render **nothing**, while a near-no-op fold (90–99% remaining) renders the brightest, nearly-complete white frame (`rgba(255,255,255,0.9)`). So aggressive folds are invisible at a glance and trivial folds shout — and a band-5 bracket set visually approximates the **pinned** tile's inset white ring, two white near-rings with opposite meanings on the same grid. Also mildly in tension with "live = solid / folded = recessed" (a 0.9-alpha white frame can make a folded tile brighter than live neighbors). Per CLAUDE.md this is yours to judge by screenshot — just flagging it.

### Also worth a sign-off
- **[MINOR] Tooltip regression** (`24476a3`): folded tooltips collapse to `kind · tok · % remaining`, dropping the "double-click to unfold" affordance hint, the toolName, and the drop-group "the agent does not see this block" + member count. Unfold discoverability via hover goes to zero — intentional per the commit, but confirm you want that.
- **[NIT]** `Inspector.svelte:145-151` has two adjacent `{#if gSavedTok > 0}` blocks (merge them); the 4-corner geometry is written twice (`tileDraw.ts drawBandBracket` and `ContextMap.svelte erosionLines`) — drift risk, and finding #2 is that drift already happening.

### Checks that passed
Math (divide-by-zero guarded, clamp kills negative/oversized so no negative badge is possible, whole-percent rounding, thorough boundary tests) · perf (sprite pattern mirrors the hatch/dice convention exactly, `dpr` plumbed, brackets inset so nothing outsets/clips, DOM SVG has `pointer-events: none` so the scroll optimization holds) · brand (text badges `--muted`/Plex Mono, no `#044EFF` misuse) · read-only rule (badges are pure store reads, identical in every mode) · Svelte (`{@const}` placement correct, no new `svelte-ignore`).

### Staleness
Branch is ~48 commits behind `devmain` but `git merge-tree` merges **cleanly** — the only shared file is `Inspector.svelte`, no textual conflict. The "41 files, 809 tests" claim is stale twice over (branch head has 51 test files; devmain now has 57 — seven suites added since). Rebase and re-run verification before the screenshot pass so the erosion-border design (finding 3) is judged against current devmain.

---

---

PR #50, opened 2026-07-03
Support multi-session discovery in browser-served mode

a-Fig:
## Summary

Fixes a user report: with no desktop app, every `pi` session opened its own browser URL on its own port, with no way to see all sessions from one page.

- `extension/accordion.ts` adds a token-gated `GET /__accordion/sessions` endpoint. The extension process is plain Node (unlike a browser tab, it's never filesystem-sandboxed), so it can read `~/.accordion/sessions/*.json` directly — the same registry the desktop app's Tauri layer reads, just from the other side. It's async (`fs.promises`, not `fs.*Sync`) so it doesn't jitter the `context` hook's 250ms fold-plan deadline under repeated polling, and it opportunistically reaps stale/dead sibling entries, since a browser-only user has no desktop app ever running to do that cleanup.
- `app/src/lib/live/browserDiscovery.svelte.ts` (new) polls that endpoint the same way the desktop app polls `list_sessions`, feeding the same reactive `discovery.sessions` state `discovery.svelte.ts` already exports — so the sidebar and session-select plumbing need no separate code path for browser-served vs. desktop. Because that poll is a relative fetch pinned to whichever session's origin served the page, a `connectedFallback()` keeps the *actively connected* session visible/reconnectable even if that original session exits (documented as a known limitation — new sibling sessions won't be discovered from that tab until a reload from a still-live session's URL).
- `SessionsSidebar.svelte`'s `browserServed` prop now only hides the pi/Claude-Code source switcher (CC transcript browsing still needs the Tauri Rust layer to read `~/.claude`); it no longer trims the session list to a single "reconnect" row.
- Rebased onto latest `devmain` to pick up the concurrently-merged remote-binding feature (#47); `selectAndConnect` forwards host/token consistently with that feature's `connectLive` signature.

Went through an adversarial subagent review before opening this PR; its findings (stale-discovery fallback, sync-fs jitter risk, missing reap, an inaccurate code comment, thin test coverage, and an explicit callout of the token blast-radius tradeoff this endpoint introduces) are all addressed in the commits.

## Test plan

- [x] `cd extension && node smoke.mjs` — includes new coverage for token-gating, cross-session listing, corrupt-sibling-file handling, stale-heartbeat exclusion, and reap-from-disk
- [x] `cd app && npx svelte-kit sync && npx svelte-check --tsconfig ./tsconfig.json` — 0 errors, 0 warnings
- [x] `cd app && npm run test -- --run` — 797 tests passing
- [ ] Manual: two `pi` sessions running the extension, open the browser URL from one, confirm both appear in the sidebar and clicking switches the live view

---
_Generated by [Claude Code](https://claude.ai/code/session_01PNHiRTsWtwV2qqcMp6FvwS)_


a-Fig:
## Recommendation: **Merge** (optional follow-ups, none blocking)

Verified in a clean worktree checkout of this branch: `smoke.mjs` passes (the new token-gating / cross-session / stale-reap / corrupt-file assertions all run and pass), `svelte-check` is 0 errors / 0 warnings across 430 files, and vitest is 797/797 with the golden snapshots byte-identical. I traced the paths that usually break in a change like this and did not find a blocking bug. The cookie-auth path is sound — the browser poll sends no `?token=` and relies on the HttpOnly `accordion_token` cookie, which `isWebAuthed` accepts, so discovery authenticates on a plain reload. `startDiscovery` correctly no-ops outside Tauri, so it never races `startBrowserDiscovery` over the shared `discovery.sessions` state. The Demo path is safe because `selectDemo` disconnects first, which breaks the new `$effect`'s `connected` guard, so Demo doesn't snap back to the live session, and that effect doesn't read `discovery.selected` so there's no reactive loop.

The thing I'd most want a second look at is the remote/off-loopback story: the sidebar now lists every sibling session, but `selectAndConnect` reuses *this page's own* token for all of them, and a genuinely remote sibling rejects that token — so those extra rows become dead clicks with a misleading "could not reach pi on :port" error. The comment flags this as a known limitation and it's honest, but from a user's chair the rail is showing sessions it can't actually open, which reads like a bug even though it isn't one. On loopback (the common case) everything works because `verifyWsUpgrade` ignores the token for a loopback peer, so I'm comfortable shipping it — I'd just want either a clearer "loopback-only switching" affordance or a distinct error state as a fast follow. The other thing worth watching is reap cadence: `listLiveSessions` reaps any non-live sibling file it reads, and now every browser tab does that every second, versus the desktop's single reaper. A live-but-briefly-wedged session (heartbeat past the 15s stale window) will have its registry file unlinked and disappear from discovery until its next heartbeat rewrites it. That mirrors the existing desktop reap semantics and the 15s threshold makes it genuinely low-risk, so it's fine, but the per-tab-per-second aggressiveness is new behavior and I'd keep it in mind if stale-flapping ever shows up in the wild.

On complexity, this PR trends the right way: it *deletes* the browser-served special-case in `SessionsSidebar.svelte` (−83/+14) and folds browser mode into the same multi-session path desktop already uses, rather than bolting on a parallel code path. The one real DRY nit is that `browserDiscovery.poll` and `discovery.poll` now carry an identical publish-and-reap tail (the `sameSessions` assign, `discovery.ready`, and the selected-gone → `disconnectLive()` block) — worth extracting a shared `publishSessions(live)` so the reap invariant can't drift between the two. Minor: `credentials: "same-origin"` is already the fetch default and the client re-runs `isLiveEntry` on a list the server just filtered and reaped, so both could go. The module banner and inline comments are heavy (~45 of the 110 new lines), and the relative-fetch limitation is documented three separate times — one authoritative spot would do. None of that is worth holding the merge for. Nice, well-tested change with genuinely honest self-documentation of its own tradeoffs.

a-Fig:
Thanks for the thorough review. Pushed a follow-up commit (2bf7b01) applying the mechanical nits:

- Extracted a shared `publishSessions()` in `discovery.svelte.ts` so the desktop and browser-served poll loops can't drift on the publish/reap-selection tail.
- Dropped the redundant `credentials: "same-origin"` (already `fetch()`'s default) and the client-side `isLiveEntry` re-filter in `browserDiscovery.svelte.ts` — the server already validates and reaps before responding.
- Consolidated the "relative fetch is pinned to the serving origin" limitation into one authoritative comment instead of three overlapping ones.

Re-ran the full verification (smoke test, `svelte-check`, 797-test vitest suite) — all green.

Left the remote/off-loopback "dead click" UX as-is for now, per your own read that it's a fast-follow rather than a blocker — that one's more of a product call (dedicated error state vs. hiding unreachable rows) than a mechanical fix, so I'd want the repo owner's steer on the right affordance before touching it.

---

PR #56, opened 2026-07-06
feat(conductors): birth-fold (#43) + recall command (#46)

### LINKED ISSUE #43: Allow conductors to fold tool_result blocks before they enters the context window

### LINKED ISSUE #46: Make recall available to conductors. Unfolding generates a cache miss, Recall does not.
Recall already exists as a command for agents. 
Conductors should be able to use it too. This could allow conductors to bring back relevant context more often without worrying about increased inference costs from cache misses.

a-Fig:
Closes #43. Closes #46.

## Why

Both issues extend what a conductor can do without paying prompt-cache misses:

- **#43 birth-fold** — a fresh oversized `tool_result` is born inside the protected working tail, so any conductor without the `tail-size` lock was clamped `protected` and the model saw the block at full size on its first call. Conductors can now fold a **never-sent** foldable block pre-flight, so the model never sees it full.
- **#46 conductor recall** — unfolding substitutes full text back mid-history → cache miss. Conductors now get the agent-recall analog: the folded block **stays folded** (digest stays in the prefix) and its original text is appended once as a synthetic user message at a **frozen anchor** near the tail — cache-safe re-injection.

## What

**Birth-fold (`4b4aacc`)**
- Sticky host-recorded exemption: `birthFolded` set + `sentThroughOrder` cursor live **outside** the per-pass reset (a fresh-only exemption would break under reset-to-baseline reapply).
- `SyncMessage.planned` (PROTOCOL_VERSION 6): only the context-hook sync advances the sent cursor; view-only syncs don't. `full:true` syncs append as already-sent (see hardening below).
- `ViewBlock.fresh` for conductor authors; bulk-loaded transcripts mark everything sent, so read-only sessions never read fresh (preview stays no-more-permissive).
- Protected-tail walk-back still uses **full** tokens — the boundary never breathes.
- `birth-fold-demo` conductor, through-store test suite, ADR 0017.

**Recall (`901ef05`)**
- `RecallCommand` (CONDUCTOR_PROTOCOL_VERSION 4) → `RecallOp {id, afterId, text}` on `PlanMessage.recalls` (PROTOCOL_VERSION 7) → `applyPlan` inserts one synthetic user message after the anchor, with group-swallow fallbacks (collapsed → after summary; dropped → after last survivor; unresolvable → append at end).
- Injected tokens count toward `liveTokens`; Inspector shows a neutral "recalled to tail" badge; recall record/release/drop emit activity-feed + decision-journal events.
- `recall-demo` conductor, through-store + mapping test suites, ADR 0018.

**Adversarial hardening (`22bc63d`)** — 3 parallel reviewers over disjoint scopes; all findings fixed, notably: reconnect replay no longer reads history as fresh (protection-bypass MAJOR), recall anchor never splits a tool_call/result pair, `recalledTokens` can't double-count mid-pass, remote example conductors bumped to protocol v4.

## Design notes (deliberate calls — flag disagreement)

1. **Recalls are sticky until explicitly released** (`restore(id)`, unfold, or detach) — merely omitting the recall from a batch does NOT drop it. This is the one deliberate departure from pure full-state command semantics: dropping an injection mutates the prefix (cache miss), so opt-out must be explicit. If you prefer pure full-state consistency at the cost of cache fragility, that's the single biggest call to revisit.
2. **Anchor is frozen at the tail at recall time** (newest durable, non-grouped, non-`tool_call` block). Mid-history anchoring was rejected as riskier for provider role-rhythm.
3. **Detach clears recalls** (folds freeze, appends don't) — costs one cache miss per recall on detach.
4. **The injected user-role message rides the same same-role-adjacency footing as shipped GroupOp summaries** (ADR 0006 watch item #1). Structurally valid; recommend one live provider check with a recall active before leaning on it hard.
5. **Grouped members clamp `not-recallable`** — a group member isn't independently on the wire; group-level recall (and keel's `GroupCommand.recoverable` ask) stay follow-ups.
6. **First full sync after attach/reconnect conservatively marks all replayed history as sent** — genuinely-new blocks in that same sync lose birth-foldability for that one call (conservative beats the protection bypass it replaces).

## Verification

- `svelte-check`: 435 files, **0 errors / 0 warnings**
- `vitest`: **833/833** across 43 files — golden `conductor.builtin.test.ts` byte-identical
- `extension smoke.mjs`: PASS (all checks incl. recall tool round-trip)
- Agent recall pi tool path: **zero behavioral diff** (never-lockable, untouched)


a-Fig:
## Merge-readiness review (6-agent adversarial shred) + fixes

Full-suite verification at `22bc63d`: 833/833 vitest, svelte-check 0/0, extension smoke green, builtin golden byte-identical. The shred found **1 HIGH + 2 MED**, all in the new #43/#46 code, all fixed in `19ebabd`:

1. **HIGH — recall could split a tool_call/tool_result pair → provider 400.** `newestRecallAnchor` skipped only blocks whose *own* kind is `tool_call`; on a view-only `message_end` conduct pass mid-tool-loop it froze the anchor on the **text/thinking sibling** of an unresolved call, and `applyPlan` spliced the synthetic user message between the call's message and its result. Fixed on both sides: the anchor picker skips every block of a tool-calling message, and `applyPlan` slides interior insertions past `toolResult` messages as a wire-side backstop (also covers the group-swallow walk-back).
2. **MED — stale `birthFolded` exemption = protected-tail bypass.** A block that settled live and crossed the wire whole on a planned sync kept its exemption. `markSent()` now prunes entries whose block is not folded at apply time, making the set truthful ("in the tail AND never sent whole") — which is also why attach/detach/reset deliberately do *not* clear it (documented in ADR 0017 §3).
3. **MED — detach popped an active birth-fold back to full.** `freezeForDetach` stamped it `override:"folded"` and the detach refold's `healProtected` immediately force-unfolded it, re-blowing the budget detach exists to protect. `healProtected` now skips blocks still carrying the exemption (ADR 0017 §4).

New regression tests pin all three (anchor sibling-skip + per-message gate, wire pairing-adjacency assertion, settle-live prune / detach freeze / conductor-swap legality). After fixes: **839/839, svelte-check 0/0, smoke green.**

Cleared by review: planned/full cursor semantics, v6/v7 skew handling, all bundled conductors bumped to v4 in lockstep, recall token accounting, no duplicate injection across passes/reconnects, recall never lockable, read-only sessions never read `fresh`.

**Notes for merge:** conflicts with #55 on `conductors/index.ts` + `conductors/README.md` (trivial, union both sides) and an ADR-number collision (both PRs claim 0017) — whichever lands second renumbers; #55's test fixture also needs `fresh: false` added once this merges. Release note: the strict v4 hello gate refuses external conductors still speaking v3. One follow-up worth a live check: a recall injects a user-role message adjacent to a user-role anchor in the common case (disclosed in ADR 0018's role-rhythm note) — one live provider run with `recall-demo` active would close it.

a-Fig:
## Verify pass + rebase — merge-ready

Two independent adversarial reviewers re-shredded the fix commit. Both confirmed the anchor fix, detach freeze, and settle-live prune sound and non-vacuously tested; their two real findings are fixed:

- `ecc7697` — **disarm→arm leak** (found independently by me and both reviewers): disarmed folding replies with an empty plan, so the wire is raw and every block crosses whole; `markSent({ rawWire })` now drops all birth-fold exemptions on those calls, closing the last protected-tail bypass. Pinned by test (m).
- `9a833a5` — the `-1` append-at-end recall path now lands before a trailing unpaired `tool_call` instead of after it (unreachable at the context hook; pure shape defense). Pinned by a new mapping test.

Then rebased onto devmain after #55 landed (`a89b395`): union-resolved the `conductors/index.ts`/`README.md` collisions, renumbered **birth-fold → ADR 0018, recall → ADR 0019** (handoff keeps 0017; keel's pre-existing plan-doc 0017 references untouched), and added the now-required `ViewBlock.fresh` to the handoff test fixture.

**Combined tree: 880/880 tests, svelte-check 0 errors / 0 warnings, extension smoke green. 0 behind devmain.**

Issue #60, opened 2026-07-08
Telemetry: plan-applied ack + passthrough counters — the extension can silently send unfolded context and nothing records it


a-Fig:
Follow-up from #58 (proposed fixes 4-5).

Every fallback branch in the `context` hook passes messages through UNFOLDED with zero observability: no GUI attached (`extension/accordion.ts:914`), plan round-trip timeout (`:920` via `requestPlan`'s 250ms timer at `:770-774`), epoch mismatch (`:921`), empty plan (`:923`), and `flushPending()` on reconnect/shutdown (`:270-281`). No log line, no counter, no wire signal.

The host side cannot detect it either: bellows' host telemetry counts plans SENT (`bellows/src/host/main.ts:496-506`, `plansSent`/`heldPlanReplies`), not plans APPLIED, and the wire protocol (`app/src/lib/live/protocol.ts`) has no ack/nack message type. Net effect (demonstrated in #58): a benchmark run's telemetry looks perfectly healthy while the model receives unfolded context on nearly every request.

Proposal:
- Add a `plan-applied` / `passthrough {cause}` message from extension -> host after each `context` event resolves, with cause one of: applied, timeout, no-gui, epoch-mismatch, empty-plan, flush.
- Count these in the host and surface "N of M requests went out unfolded (by cause)" in bellows' ConductorTelemetry and the run report.
- In the bellows report, cross-check provider usage (`in + cacheRead` = actual wire prompt size) against Accordion's `wireTokens` belief per turn and flag divergence — this is the alarm that would have caught #58 automatically, and it can also triage existing historical runs for contamination without re-running them.

Note: per-request plan round-trip TIME tracking (avg + per-turn in the bench transcript) is being shipped separately alongside the #58 fix; this issue covers the applied/passthrough ack and cause counters, which require a wire-protocol addition.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #59, opened 2026-07-08
Perf: linearize() re-processes the entire message array on every context event (O(session) work per model request)


a-Fig:
Follow-up from #58 (proposed fix 3).

Every `context` hook invocation runs `linearize(lastMessages)` over the ENTIRE message array (`extension/accordion.ts:911-913` -> `app/src/lib/live/mapping.ts:117-171`), calling `estTokens()` on every text/thinking/tool_call/tool_result part of every message — purely to compute the delta since `sentCount`. This is O(total session characters) of CPU on every single model request, and it shares the event loop with the WS sync replies. At 100k+ token sessions this is a prime driver of the 250ms plan-window misses documented in #58 (folds stopped landing entirely once the session passed ~115k tokens).

Host-side has the same shape: every deferred conduct runs `store.refold()` -> `buildView()` over every block (`app/src/lib/engine/store.svelte.ts:1051-1068`), though per-block digests are already WeakMap-memoized (`app/src/lib/engine/digest.ts:95-96`).

Proposal:
- Memoize linearization per message, keyed on message identity (WeakMap), so each `context` event only processes the appended tail. Messages are append-only in practice; take care with the final message if it can still be streaming/mutated.
- Consider the same treatment for the store-side full-array passes (`buildView`, `healProtected`, `clearConductorState`).

The per-block memoization pattern already exists in this codebase (`digest.ts` WeakMap caches, `code-skeleton`'s block-id cache, `keel`'s signature-keyed LLM cache) — this issue is about applying it to the extension hot path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

a-Fig:
Investigation findings (traced through pi's vendored SDK) that change the proposed fix:

**1. The WeakMap-per-message-object proposal cannot work.** pi's `emitContext` runs `structuredClone(messages)` on EVERY context emission (`runner.js:685-687` in the vendored `@earendil-works/pi-coding-agent`), so every message is a brand-new object identity each event — a WeakMap keyed on message objects has a guaranteed 0% hit rate for `lastMessages`. Any viable cache must key on the durable content-anchored `blockId` (timestamp/responseId/toolCallId — survives clone churn), the same pattern `conductors/code-skeleton` already uses with its string-keyed cache. (Plan-application mutation poisoning is a non-issue: `applyPlan` is pure and its output never feeds back into pi's canonical history.)

[...]

Issue #58, opened 2026-07-08
Silent unfolded passthrough: 250ms plan timeout + per-request folds let context blow 52% past budget while telemetry reads healthy


a-Fig:
## Summary

On a bellows benchmark run (`runs/_worker/14ea0b95b4a5bdbe/compaction-naive-0`: conductor `compaction-naive`, budget 100k, protect 20k, `deepseek/deepseek-v4-pro` via token-router), provider-reported usage proves the model received the **full, unfolded conversation on nearly every request** — actual prompt size (`in + cacheRead`) grew almost monotonically to **152,023 tokens, 52% past the 100k budget** — while Accordion's own `wireTokens` telemetry believed the folded prompt ended the run at ~63k, comfortably under budget.

The conductor did its job: `liveTokens` crossed the 90% trigger (~90k) around turn 120, summarization fired, and the plan was produced. The plan just almost never made it onto the wire.

## Evidence (from the run's provider usage vs Accordion telemetry)

- 200 assistant turns. `in + cacheRead` (which for the DeepSeek adapter is exactly per-request prompt size: cache-hit tokens → `cacheRead`, cache-miss → `in`) climbs from ~2.7k to 152k, crossing 100k at turn ~117 and never coming back down.
- Exactly **6 requests went out folded** (turns ~125–142): each shows the unmistakable signature of a rewritten prompt — large fresh `in=` (up to 31,311), tiny `cacheRead`, total ~34–43k. Each is followed **on the very next turn** by a prompt back at full unfolded size (`cacheRead` ≈ pre-fold level).
- After turn ~142 no fold ever lands again; context grows freely for the remaining ~58 turns.
- Meanwhile `record.json` shows `wireTokens` dropping 90,104 → 22,622 at the fold and ending at 63,496 — Accordion's belief and the wire diverge by 2.4×.
- `cacheWrite=0` on all turns is expected (DeepSeek reports hit/miss only; no cache-write field) — not part of the bug.

## Root cause chain

1. **The fold must be re-won on every request.** pi applies `transformContext` only to the outgoing payload of that single call (`pi-agent-core/dist/agent-loop.js:172-179`); the transformed array is never written back to session state. So every turn of every tool loop is an independent plan round-trip.
[...]

PR #61, opened 2026-07-09
fix(extension): stale-plan fallback + steering mode + plan RTT (#58)

a-Fig:
## Why

Fixes the silent-unfolded-passthrough failure documented in #58. The pi extension's `context` hook waited 250ms for the GUI's fold plan and, on timeout, resolved the **same** empty plan `{ops:[],groups:[]}` that a genuinely-empty reply produces. So "the plan was slow" was indistinguishable from "the conductor wants no folds" — and every timed-out request went out **unfolded, silently**. In a real bellows run this let context climb to 152k tokens (52% past a 100k cap) while Accordion's own telemetry read ~63k.

This PR implements the issue's **proposed fixes #1 and #2** (steering/blocking mode + apply-last-known-plan-on-timeout) and lays the **RTT groundwork** the report asks for. It does **not** touch the O(session) hot-path memoization (#3) or the ack/telemetry cross-check (#4/#5) — those are tracked separately (issues #59 / #60).

## What changed (all in `extension/accordion.ts`)

**Feature A — stale-plan fallback (always on).**
- `requestPlan` now resolves a discriminated `PlanResult` — `"plan"` (GUI replied, possibly empty) · `"timeout"` (waited, no reply) · `"unsent"` (no GUI / socket dropped). `flushPending()` and the no-GUI path keep their existing passthrough-without-advancing meaning (they resolve `"unsent"`).
- A module-level `lastPlan` caches the last **delivered** plan — **including a genuinely empty one**, because empty = "conductor wants no folds", and caching it stops a later timeout from resurrecting a stale non-empty plan. Reset on `session_start` and every GUI (re)connect (wherever the cursor/epoch resets).
- On timeout the hook re-applies `lastPlan` (id-addressed; `applyPlan` passes through ops for ids no longer present) instead of shipping unfolded, and **logs** the fallback (`cause + reqId + elapsed`). If there's no cached plan it passes through as before — but still logs. Never silent again.

**Feature B — configurable timeout + steering (blocking) mode.**
- `ACCORDION_PLAN_TIMEOUT_MS` (default 250) replaces the hardcoded constant.
- `ACCORDION_STEERING` (truthy `1`/`true`) switches the wait to a hard `ACCORDION_PLAN_DEADLINE_MS` (default 10000). A missed deadline logs **loudly** via `console.error` and still falls back to the stale plan. It never blocks when no GUI is attached (the `!attached()` guard runs first), and a mid-wait disconnect now flushes pending in the socket `drop` handler so the wait resolves immediately rather than stalling the model call for the whole deadline.
- Env values parsed defensively (`NaN` / ≤0 / non-integer → default).

**Feature C — plan RTT surfaced in the session file.**
- `context` measures the wall-clock ms it waited on the plan and stashes it; the assistant `message_end` handler stamps it onto the message as **`usage.rttMs`** (integer ms), then clears the stash (a message with no preceding context wait gets no field). `rttMs` = time awaiting the plan for the request that produced this assistant message, whatever the outcome (applied / stale-fallback / timeout / no-host ≈ none).

## Design notes for reviewers (the two verifications from the task)

- **Timeout sentinel:** rather than a magic `null` + out-of-band flag, `requestPlan` returns a small discriminated union. This is what makes "timed out" cleanly separable from "delivered empty" and "never sent" at the one call site.
- **`message_end` composition (3a):** I verified the vendored SDK — `loader.js` **appends** handlers per event (`list.push`) and `runner.js#emitMessageEnd` **chains** each handler's returned message — so a *second* `pi.on("message_end", …)` would compose fine **in production**. I still **merged** the RTT injection into the existing view-only handler because (1) `smoke.mjs`'s mock `pi.on` overwrites per event (last-wins), so a second handler would silently drop one in tests, and (2) injection must run **even when no GUI is attached**, otherwise a value stashed while a GUI was briefly attached could leak onto a later message — one handler above the no-GUI guard guarantees the stash is always consumed+cleared.
- **`usage.rttMs` persistence (3b):** verified end-to-end. `message_end` replacements are applied in place (`agent-session._replaceMessageInPlace`) and `SessionManager.appendMessage` JSON-serializes the **whole** message, so arbitrary `usage` keys survive to the session file. No `appendEntry` fallback needed — **downstream consumers can read `usage.rttMs` directly off the assistant message** (integer ms; absent when there was no plan wait for that message, e.g. no GUI attached).

## Tests

- `smoke.mjs` (default-env, in-process): timeout → **stale plan re-applied** (not passthrough); a delivered **empty plan is not overridden** by the older plan (a later timeout passes through unfolded); **`usage.rttMs`** injected on the assistant message and **cleared** so the next message gets no field.
- `smoke-config.mjs` (new): module-init env is per-process, so it spawns a child per scenario — **steering deadline** (600ms honored, `console.error` fired, stale plan applied), **invalid env → 250 default** (`console.warn`), **custom timeout honored** (120ms). Wired into `npm run smoke`.

All green:
```
SMOKE PASS — … stale-plan fallback ✓  rttMs stamp ✓
SMOKE-CONFIG PASS — steering deadline ✓  env defaults (invalid → 250) ✓  custom timeout honored ✓
```

Docs: file header + `README.md` gain a Configuration section for the env vars and the `usage.rttMs` semantics.

### DIFF
--- a/extension/README.md
+++ b/extension/README.md
@@ -72,6 +72,23 @@ reasons over at full fidelity.
 <img src="https://raw.githubusercontent.com/a-Fig/Accordion/main/docs/assets/attention-conductor.png" alt="Attention conductor view — each block tinted by how much the working tail still attends back to it" width="600">
 </div>
 
+## Configuration
+
+Advanced knobs, read from the environment once when the extension loads. Defaults are
+tuned for interactive use; the steering knobs matter mainly for benchmark harnesses that
+must enforce a hard context budget.
+
+| Env var | Default | Effect |
+|---|---|---|
+| `ACCORDION_PLAN_TIMEOUT_MS` | `250` | How long a model request waits for the GUI's fold plan before falling back. On a miss the extension re-applies the **last known plan** rather than shipping the conversation unfolded (a one-turn-stale plan is strictly better than none), and logs the fallback — it is never silent. |
+| `ACCORDION_STEERING` | off | Truthy (`1`/`true`) switches the wait to a hard **deadline** (`ACCORDION_PLAN_DEADLINE_MS`) instead of the short timeout, so a run whose cap must hold actually holds it. A missed deadline is logged loudly and still falls back to the last known plan. It never blocks when no GUI is attached, and a mid-wait disconnect resolves immediately. |
+| `ACCORDION_PLAN_DEADLINE_MS` | `10000` | Steering-mode deadline. Ignored unless `ACCORDION_STEERING` is on. |
+
+Invalid values (non-numeric, `NaN`, `≤0`) fall back to the default.
+
+Each request also records the plan round-trip time it waited, stamped onto the assistant
+message it produced as `usage.rttMs` (integer milliseconds) in the session file.
+
 ## Skills included
 
 This package registers two pi skills the agent uses to interact with folded context:
--- a/extension/README.md
+++ b/extension/README.md
@@ -81,10 +81,12 @@ must enforce a hard context budget.
 | Env var | Default | Effect |
 |---|---|---|
 | `ACCORDION_PLAN_TIMEOUT_MS` | `250` | How long a model request waits for the GUI's fold plan before falling back. On a miss the extension re-applies the **last known plan** rather than shipping the conversation unfolded (a one-turn-stale plan is strictly better than none), and logs the fallback — it is never silent. |
-| `ACCORDION_STEERING` | off | Truthy (`1`/`true`) switches the wait to a hard **deadline** (`ACCORDION_PLAN_DEADLINE_MS`) instead of the short timeout, so a run whose cap must hold actually holds it. A missed deadline is logged loudly and still falls back to the last known plan. It never blocks when no GUI is attached, and a mid-wait disconnect resolves immediately. |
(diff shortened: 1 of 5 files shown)


a-Fig:
Pushed a follow-up commit ([\`a3c5514\`](https://github.com/a-Fig/Accordion/commit/a3c5514)) addressing every finding from the adversarial review:

**BLOCKER — usage.rttMs never persisted.** The `message_end` handler was returning a bare `AgentMessage` at all three exit points. pi's `MessageEndEventResult` contract requires `{ message }` — `emitMessageEnd` does `if (!handlerResult?.message) continue;` — so the replacement (and the rttMs stamp on it) was silently dropped every time. All three exits now go through one `finish()` helper that wraps correctly.

**SHOULD-FIX — stale RTT stash mis-attribution.** `lastPlanRttMs = null` now runs at the very top of the `context` hook, before the no-GUI early return, so an aborted turn followed by a GUI detach can no longer leak an old RTT onto an unrelated message.

**SHOULD-FIX — session_start didn't invalidate in-flight plan waits.** `session_start` now calls `flushPending()` and bumps `epoch`, mirroring the GUI-reconnect path, so a plan for the OLD session landing after a mid-flight session switch fails the epoch guard instead of corrupting the new session's cursor/`lastPlan`.

**Nits:** the timeout log now distinguishes "cached empty plan applied" from "no cached plan at all"; the header/README document that `envPositiveInt` rejects non-integers (`"250.5"`) but accepts `Number()`-parseable forms like `"1e3"`/`"0x10"`; the header now notes that a timeout-with-stale-fallback turn can leave the GUI's view diverged from what was actually sent for one turn (invisible to the fold alarm) until the next delivered plan resyncs.

**Doc-only finding:** README now documents that `ACCORDION_STEERING` has no circuit breaker yet — a hung-but-connected GUI stalls every model request for the full deadline — and that's tracked as follow-up work only, not implemented here.

**Test harness fix:** `smoke.mjs`'s stale-plan/rttMs scenario previously read `.usage` straight off the raw `message_end` return, which is exactly the shape the blocker broke — so the test would have passed even with the bug. It now unwraps via the same `{ message }` contract the real pi runner enforces. Verified by temporarily reverting the `{ message: replacement }` wrapper: `npm run smoke` failed on both rttMs assertions as expected, then passed again once restored.

Verification: `npm run build:extension` clean, `npm run smoke` fully green (including `smoke-config.mjs`).

PR #72, opened 2026-07-10
fix(extension): harden loopback WS auth and completion relay

a-Fig:
## Why

Post-#81 Accordion is loopback-only, but loopback alone does not stop a hostile website from opening a WebSocket to `127.0.0.1`: WebSockets do not enforce CORS. A browser page that found the ephemeral port could replace the active GUI, read the backlog, steer fold plans, and trigger billable out-of-band completions. The completion relay also had no concurrency or lifetime bound.

This branch has been rebuilt on current `devmain` after #81. The removed remote-binding, birth-folding, and conductor-recall code is not reintroduced.

### WebSocket authorization

- Keep the server bound to `127.0.0.1` on an ephemeral port and reject non-loopback peers again at upgrade time.
- Accept tokenless native clients with no browser `Origin` and the fixed Tauri v2 custom-protocol origins used by production builds; gate the browser-obtainable Vite development Origin behind an explicit environment flag.
- Accept browser clients with the explicit per-session bearer from `/accordion`.
- Accept the ambient HttpOnly cookie only when `Origin` exactly matches `Host` on a literal loopback hostname. This closes cross-port cookie replay and DNS-rebinding authorization.
- Preserve browser multi-session switching by verifying a sibling literal-loopback Origin through a live `/__accordion/meta` probe plus matching live registry identity.
- Bound sibling probes by concurrency (8), deadline (750 ms), response size (16 KiB), loopback peer, JSON content type, protocol, session id, and port.
- Parse malformed upgrade targets inside `try/catch` and reject them with 400.
- Cap inbound WebSocket frames at 8 MiB.

### Completion relay

- Limit billable `completeRequest` work to four concurrent calls across socket reconnects.
- Use unique call tokens so an old call cannot release a newer call that reused its request id.
- Scope duplicate request-id coalescing to one connection; a duplicate no longer rejects the original correlated promise.
- Reject unsafe request ids and invalid `maxOutputTokens` values.
- Apply a 110-second client deadline and pass an `AbortSignal` to pi-ai.
- If a provider ignores abort, keep its global slot occupied until the underlying provider promise actually settles. The timeout does not falsely reopen spend capacity.

## Regression coverage

`extension/smoke.mjs` drives real HTTP upgrades and WebSocket frames for:

- hostile and opaque browser Origins;
- native and Tauri origins;
- explicit bearer and exact-origin cookie paths;
- DNS rebinding and cross-port cookie replay;
- live, stale, non-ending, forged, and oversized sibling-Origin probes;
- malformed targets and oversized frames;
- duplicate completions, the four-call ceiling, reconnect preservation, pre-provider watchdog recovery, and a provider that ignores abort.

## Verification

- `npm test` — 927/927 tests
- `npm run check` — 0 errors / 0 warnings
- `npm run smoke` — pass, including the new real-socket security cases
- `npm run build:extension` — pass
- `npm pack --dry-run` — pass; tarball contains the rebuilt `accordion.js`, browser client, and both skills
- `git diff --check` — clean
- Two adversarial review passes after the post-#81 rewrite; the provider-settlement accounting finding was fixed and regression-tested

## Platform scope

The production allowlist matches Tauri v2's documented custom-protocol origins (`tauri://localhost`, `https://tauri.localhost`, with the legacy HTTP WebView2 variant). The browser-obtainable Vite development Origin (`http://localhost:1420`) is rejected by default and requires `ACCORDION_ALLOW_TAURI_DEV_ORIGIN=1` in the pi process; both default rejection and opted-in acceptance are covered in smoke tests. macOS WKWebView behavior is out of scope for this change and is not a merge gate; `Origin: null` intentionally remains rejected because ordinary sandboxed browser documents can also produce it.


a-Fig:
Static review of `50ef423` against current `devmain` (`c5cf994`). The malformed-target guard and `maxPayload` cap look correct, but I found four issues that should be addressed before merge:

1. **[P1] Generic same-origin authorization is vulnerable to DNS rebinding.** `isSameOriginUpgrade()` trusts `Origin.host === Host`. A page served from `http://evil.example:P` can rebind that hostname to `127.0.0.1`, open `ws://evil.example:P`, and arrive with matching Origin/Host plus a loopback peer address, so it is accepted tokenless. The served UI already has a query token or cookie; remove the generic tokenless same-origin exemption, or constrain it to normalized literal loopback hosts.

2. **[P1] Cookie authentication crosses ports.** Port-qualified cookie names prevent sibling-session clobbering, but cookies remain host-scoped. A hostile page on the same hostname at another port automatically sends Accordion's HttpOnly cookie when opening the Accordion WS port. Because query-token and cookie auth are combined into `hasToken`, that cross-origin page is accepted. Treat them separately: an explicit query token may override Origin, while cookie auth should require a validated same-origin request.

3. **[P1] Reconnects bypass the completion ceiling.** Every new connection calls `inFlightCompletions.clear()` while old provider calls keep running. Launch four, reconnect, launch four more, and repeat: actual concurrent paid calls are unbounded. Reusing an old reqId is worse—the old call's `finally { delete(reqId) }` can remove the new call's slot. Keep global active-call tokens until their exact upstream calls settle; track request IDs separately per connection/epoch.

4. **[P2] Duplicate rejection settles the original caller.** Sending `{ok:false}` under the duplicate's same `reqId` rejects/deletes the GUI's original pending promise even though its paid provider call continues and its eventual success is ignored. Ignore/coalesce an exact in-flight duplicate, or use collision signaling that cannot settle the original request.

Please add regression coverage for matching arbitrary Origin/Host (DNS rebinding), a different-port Origin carrying the ambient cookie, reconnect while four provider calls are parked, and reqId reuse across connection generations.

Additional watch items: prompt/system can still approach 8 MiB each, provider work has no extension-side timeout/cancellation, `maxOutputTokens` needs finite positive-integer validation and a hard fallback ceiling, and the nominally per-session token/client authorization remains unchanged across later `session_start` events.

Static review only; I did not build or run live tests.

a-Fig:
## Adversarial security review — WS auth hardening

Traced the full `verifyWsUpgrade` path against `origin/fix/ws-auth-hardening`. Net improvement over the prior fully-tokenless loopback WS, and the crash-safe verifier + completion bounds are correct. But this should **not** merge as the definitive CSWSH fix — the loopback defense is still bypassable two ways, both of which are exactly what #73 sets out to fix.

### What's solid
- **Fix #1 (crash-safe verifier) — correct.** The `new URL(req.url…)` parse is inside try/catch on the single path through `verifyWsUpgrade`, and the two other pre-auth parses (`isSameOriginUpgrade`, `isTrustedNativeWebviewOrigin`) each have their own. No throw-capable code remains reachable pre-auth; malformed target → `cb(false, 400)`.
- **Completion bounds — race-free, no slot leak.** `.add` is immediately followed by `try { … } finally { delete }`; the three early rejections `return` before `.add`; the dup/ceiling checks + `.add` run synchronously before the first `await`. The concurrency claim holds within a connection generation.
- **No context-hook / blocking-I/O violations.** All changes live in the upgrade path, the `WebSocketServer` config, and the out-of-band completion handler.
- **Smoke tests are non-vacuous** — the malformed-target, Origin-split, 9 MiB-frame, and completion dup/ceiling assertions all check real behavior.

### Confirmed findings

**1. [MAJOR] Same-origin tokenless exception is DNS-rebinding vulnerable** — `isSameOriginUpgrade` (L1063–1066) returns true whenever `new URL(origin).host === req.headers.host`, with **no requirement that the hostname be a loopback literal**. Attacker serves `http://a.attacker.com:P`, DNS-rebinds it to `127.0.0.1`; post-rebind the peer is loopback (`isLoopbackPeer` true) and `Origin`/`Host` both read `a.attacker.com:P` → tokenless admission at L1098. The attacker then becomes the active client, submits plans, toggles armed, and fires paid completions. This is #73's "same-origin is DNS-rebinding vulnerable" criticism, and it is real in this branch — the fix is #73's added loopback-literal-hostname requirement.

**2. [MAJOR] Ambient HttpOnly cookie re-opens loopback CSWSH across ports** — `hasAccordionCookie` (L1052) feeds `hasToken` (L1085). The cookie is `SameSite=Strict; Path=/`, but SameSite does **not** isolate by port. A malicious page at `http://127.0.0.1:3000` making a WS request to `127.0.0.1:<accordionPort>` is same-site, so the browser auto-attaches `accordion_token_p<port>` if the user ever opened the served UI → admitted via the ambient cookie, without the attacker ever reading the token. The port-qualified cookie *name* stops a *different* session's cookie from matching but not the *correct* cookie being ambiently replayed. Also #73's criticism, also confirmed.

**3. [MAJOR / needs live verification] macOS/Linux desktop dial may be locked out by `Origin: null`** — the Tauri app dials tokenless (`liveClient.svelte.ts:263`, `ws://127.0.0.1:port`, no token). If WKWebView/WebKitGTK attaches a literal `Origin: null` on the WS handshake, `origin === "null"` is a non-empty string, `new URL("null")` throws in both predicates → `cb(false, 403)`, and the app can't steer its own session — a regression of the CLAUDE.md "tokenless dial is unaffected" invariant. Header-*omitted* works; `Origin: null` breaks. The smoke test can't cover the real webview — **must be verified on native macOS/Linux builds before merge** (your own description flags this).

### Minor / nit
- **[MINOR] Completion counter corrupts across connection generations** — `inFlightCompletions` is keyed by bare `reqId`; `clear()` on a new connection (L1179) plus a parked older completion's `finally delete(reqId)` (L1340) can delete a newer client's live entry. Bounded to ≤4 leaked slots (~8 concurrent worst case); self-acknowledged.
- **[NIT] `isTrustedNativeWebviewOrigin` over-permissive** (L157–172): `tauri:` accepts any host; `hostname === "tauri.localhost"` accepts any scheme/port. Not browser-forgeable, so no exploit — just wider than "a small fixed family."

### Bottom line
Merge #1 and the completion bounds happily. But findings #1 and #2 mean the CSWSH story isn't closed here — that's what #73 exists for, so I'd either fold #73 in or land this explicitly as "crash-fix + bounds, CSWSH tracked in #73." Verify finding #3 on a real macOS build first regardless, since it can break the desktop dial.

---

---

a-Fig:
Rebuilt this branch from current post-#81 `devmain` and force-updated the PR. The old remote-binding/birth-fold/conductor-recall code is gone. The retained fixes now target concrete loopback CSWSH, cross-port cookie/DNS-rebinding authorization, bounded sibling verification, frame limits, and completion spend/concurrency races. I also fixed the abort-ignoring-provider accounting gap found during the final adversarial pass and added a regression that proves slots stay occupied until the real provider promise settles. Full validation and the remaining native platform check are documented in the updated PR body.

a-Fig:
## Review — solid design, one thing to fix before merge

Reviewed `ead03c6` against current devmain: read the auth path in full plus two independent adversarial passes (WS-upgrade logic; completion caps + full verification suite). Verdict: the hardening is fundamentally sound and the CSWSH defense holds. **One MAJOR to fix, one item to verify.** The whole design correctly rests on the one true property — a browser cannot forge `Origin` — and the choice to preserve sibling-session switching via a bounded `/meta` + registry probe (rather than dropping the feature) is the right call and is implemented carefully (DNS-rebind and cross-port-cookie cases are both closed).

### MAJOR — drop the unconditional `:1420` dev-origin trust
`extension/accordion.ts:162` (`isTrustedTauriOrigin`): `http://localhost:1420` / `http://127.0.0.1:1420` are accepted on the Origin string **alone** — no token, no cookie, no port-ownership probe. Unlike every other accept path, this trusts a fixed well-known port unconditionally.

- **It protects no real shipped client.** An npm install has no Vite dev server; `npm run tauri dev` connects via the `tauri://` / `tauri.localhost` origins, not `:1420`; `npm run dev` is UI-only (no live link). With the desktop app slated to be dropped (#79) it's even more clearly dead.
- **It only widens the exact surface this PR closes.** Any browser-driven foothold on 1420 (a stale service worker from a past dev run, reflected content in another local app that lands on 1420) yields a tokenless steering+paid-completion socket.

Fix: remove the `:1420` clause (dev can pass `?token=`), or gate it behind an explicit dev/env flag so it never ships enabled. Everything else in `isTrustedTauriOrigin` stays.

### VERIFY (not a code defect) — macOS WKWebView `Origin: null`
If Tauri v2's macOS webview emits `Origin: null` on its own handshake, the desktop app self-connect currently 403s (`new URL("null")` throws → all paths fail). Windows WebView2 (`https://tauri.localhost`) and dev are covered; macOS is unverified statically. Confirm with a live `npm run tauri dev` connect on macOS before relying on it; if it's `null`, add a scoped allow (combined with the existing loopback-peer gate so a sandboxed-iframe `null` can't abuse it). Given #79 this may become moot.

### Verified safe / correct (checked, for the record)
- CSWSH: no browser-obtainable Origin passes any accept path except the `:1420` gap above. No-Origin native, bearer, cookie+exact-origin, and sibling-probe paths all hold.
- Sibling `/meta` probe: constrained to loopback GET, 750ms / 16KiB / 8-concurrent caps, boolean result; the sessionId+port registry cross-check is a real gate a browser attacker can't satisfy. `pendingSiblingOriginProbes` decrements on every termination path (timeout/abort/throw/parse-fail) — no counter wedge.
- `isExactServedOrigin` fail-closed against punycode, trailing-dot, IPv6 forms, uppercase, embedded creds, multi-value Host.
- Completion caps: `MAX_CONCURRENT_COMPLETIONS=4` process-wide, counter always released in `finally`, survives socket supersede, timeout threads a real AbortSignal into the provider; `maxPayload` bounds frames before parse; relay stays out-of-band (never on the context hook).
- smoke.mjs (+261): genuine end-to-end coverage of the reject/accept/bounds paths — not assertion theater.
- Suite green on `ead03c6`: svelte-check 0/0 (438 files), vitest 927/927, smoke PASS, `npm pack --dry-run` tarball complete. PROTOCOL_VERSION unchanged (9) both sides; no reintroduction of removed features.

Fix the `:1420` trust and (owner) settle the macOS check, and this is merge-ready.

a-Fig:
Addressed the `:1420` review finding in `0625c1a`. The Vite/Tauri development Origin is now rejected in shipped/default mode and accepted only when the pi process explicitly sets `ACCORDION_ALLOW_TAURI_DEV_ORIGIN=1`. Production Tauri custom-protocol origins remain trusted. Added real-handshake coverage for default rejection plus a separate-process config scenario for opted-in acceptance; `npm run smoke` and `npm run check` both pass.

PR #73, opened 2026-07-10
fix(extension): close PR #72 WS auth and limiter gaps

a-Fig:
## What changed

This follow-up targets #72's `fix/ws-auth-hardening` branch and addresses the issues from the static review on that PR.

- Removes generic tokenless `Origin.host === Host` authorization, which was vulnerable to DNS rebinding.
- Separates explicit query-token authority from ambient cookie authority. A cookie is accepted only for its exact served Origin, so unrelated same-host/different-port pages cannot inherit WS access.
- Preserves loopback browser multi-session switching without trusting arbitrary ports: a tokenless sibling Origin must be a literal loopback address, answer the live Accordion `/__accordion/meta` contract, and match a current live registry `sessionId` + port.
- Bounds sibling-Origin verification with an absolute 750 ms deadline, a 16 KiB byte cap, immediate response teardown on failure, and an eight-probe global ceiling.
- Makes paid-completion accounting global across reconnects using unique call tokens. Request-id deduplication is scoped per connection, and duplicate replays coalesce onto the original result instead of rejecting its promise.
- Rejects non-safe-integer completion request IDs.

## Why

#72's original same-origin exception allowed matching attacker-controlled Origin/Host values after DNS rebinding. Its host-scoped cookie was also accepted from hostile pages on other ports. Separately, reconnecting cleared completion bookkeeping while upstream provider calls continued, allowing the four-call spend ceiling to be bypassed and old `finally` cleanup to erase newer slots that reused the same numeric request ID.

The loopback sibling proof is deliberately two-part. A live `/meta` response proves current ownership of the exact Origin port; the registry identity match prevents an unrelated local HTTP service from authorizing itself by imitating the public JSON shape. Same-user native processes remain inside the existing intentional no-Origin trust boundary.

## Regression coverage added

`extension/smoke.mjs` now covers:

- DNS-rebound matching Origin/Host rejection;
- hostile cross-port cookie rejection;
- forged `/meta` without registry identity rejection;
- live sibling Accordion Origin acceptance and immediate rejection after shutdown/removal;
- non-ending failed-probe teardown;
- exact-origin cookie and explicit-token success paths;
- duplicate completion coalescing with one correlated result;
- reconnect preserving the global completion ceiling;
- capacity reopening only after the actual provider promises settle.

## Validation

- `node --check extension/smoke.mjs`
- `git diff --check`
- Independent static review of the final auth and completion-accounting paths

Per the requested review constraint, I did **not** build the app, run the smoke suite, or perform live Pi/Tauri testing.

### DIFF
--- a/extension/accordion.ts
+++ b/extension/accordion.ts
@@ -133,6 +133,39 @@ const UNFOLD_TIMEOUT_MS = 2000;
 // folded block's full content back THIS turn — so the same generous wait applies.
 const RECALL_TIMEOUT_MS = 2000;
 
+// ── WS hardening (private security review of the devmain promotion) ──────────────
+// Cap on any single inbound WS frame. `ws` defaults to 100 MiB (maxPayload), which lets an
+// unauthenticated loopback peer stream ~100 MiB per message straight into JSON.parse. Real
+// plans/backlogs are kilobytes; 8 MiB is a comfortable ceiling far below the default.
+const MAX_WS_PAYLOAD_BYTES = 8 * 1024 * 1024;
+// Cap on concurrently in-flight paid provider completions (`completeRequest`). Each accepted
+// request launches a real, billable model call; without a ceiling a client (or a hijacked
+// socket) can fan out unbounded spend. The legitimate handoff/compaction path issues one at a
+// time, so a small cap never blocks it.
+const MAX_CONCURRENT_COMPLETIONS = 4;
+// Does this browser Origin belong to the trusted Tauri DESKTOP webview? The Tauri app dials
+// `new WebSocket("ws://127.0.0.1:<port>")` from inside its own webview (see
+// app/src/lib/live/liveClient.svelte.ts) — so WebView2 (Windows) / WKWebView (macOS/Linux)
+// DO attach an Origin to the handshake. NOTE: this contradicts the "Tauri sends no Origin"
+// assumption — the dial is a browser WebSocket, so it always carries one. The webview's origin
+// is one of a small fixed family (see tauri.conf.json: devUrl http://localhost:1420;
+// frontendDist ⇒ the `tauri.localhost` custom-protocol origin in a production build). A hostile
+// web page in the user's NORMAL browser cannot forge any of these: the browser sets Origin from
+// the page's real origin, and no public site is served from the `tauri:` scheme or the
+// `tauri.localhost` host. Matched by predicate (not an exact-string set) so a scheme/port
+// variant across Tauri/OS versions still authorizes the legitimate desktop dial.
+function isTrustedNativeWebviewOrigin(origin: string): boolean {
+	let u: URL;
+	try { u = new URL(origin); } catch { return false; }
+	// macOS / Linux custom protocol: tauri://localhost
+	if (u.protocol === "tauri:") return true;
+	// Windows (WebView2) custom protocol: http(s)://tauri.localhost
+	if (u.hostname === "tauri.localhost") return true;
+	// `npm run tauri dev`: the webview loads devUrl http://localhost:1420
+	if ((u.hostname === "localhost" || u.hostname === "127.0.0.1") && u.port === "1420") return true;
+	return false;
+}
+
 // Base dir is overridable for tests (smoke.mjs) so they don't touch the real home.
 const HOME = process.env.ACCORDION_HOME || os.homedir();
 const REGISTRY_ROOT = path.join(HOME, REGISTRY_DIR);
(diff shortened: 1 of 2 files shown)


a-Fig:
## Review — auth follow-up (closing #72's gaps)

Traced the full diff (`fix/ws-auth-hardening...agent/ws-auth-followup`), the head file, and the app-side dial path. **The security logic is sound and answers every adversarial angle I threw at it** — the rebinding fix, exact-origin cookie scoping, per-connection dedup, and probe containment are all correctly constructed. Two things stand between draft and ready.

### Confirmed the fixes land
- **DNS-rebinding closed** — `isCookieAuthorizedUpgrade` (~L1072) now additionally requires a literal loopback hostname, so #72's `isSameOriginUpgrade` bypass is shut.
- **Cross-connection duplicate-result leak — none.** `connectionCompletionIds` is declared inside `wss.on("connection")` (~L1294), so dedup is per-connection; a new connection replaying an old reqId is a fresh billable request, and the original result can't reach it (`reply` is closed over `capturedWs` and only sends when `capturedWs === client`).
- **SSRF well contained** — probe target is restricted to literal `localhost`/`127.0.0.1`/`[::1]` before any I/O, fixed path/method, 16 KiB/750 ms caps. A browser attacker can only make Accordion probe *its own* port (browser pins Origin); forging an arbitrary-port Origin requires a raw local client already inside the tokenless native trust boundary.
- **Browser-served UI (the primary npm flow) survives, belt-and-suspenders** — the served page dials tokenless (`liveClient.svelte.ts:224`), the browser attaches the port-qualified `SameSite=Strict` cookie → `isCookieAuthorizedUpgrade` passes; if the cookie is absent, the sibling probe self-probes this very server (`/__accordion/meta` + registry `sessionId+port` match) and still admits. A→B multi-session switch works the same way.
- **No blocking I/O on protected paths** — the probe and registry read run only inside `verifyWsUpgrade`, never on the `context` hook; `listLiveSessions` is async `fs.promises` with best-effort swallowing. CLAUDE.md invariants hold.

### Findings

**1. [MAJOR] Hung provider calls now permanently consume global completion slots — reconnect is no longer a recovery path.** Admission gates on `activeCompletionCalls.size >= 4` and the slot frees only in `finally` on settlement. Neither `getApiKeyAndHeaders` (~L1414) nor `complete(...)` (~L1433) has a deadline. Removing the old per-connection `clear()` fixed the billing bypass but also removed the only recovery: if 4 provider calls wedge (network blackhole / provider hang), completions are bricked for the rest of the pi session with no user recourse. To be clear — the cap is **4 in-flight, not 4-total-forever** (a settled call reopens capacity, and the new smoke asserts this), so legitimate long sessions are fine; the regression is confined to never-settling calls. **Verdict: PLAUSIBLE** (real if `complete()` can hang). Fix: wrap each provider call in a hard deadline (`Promise.race`) so the token always frees.

**2. [MAJOR / process] None of this has ever executed.** ~90 lines of new smoke assertions plus the security logic; `node --check` proves syntax only. I found no undefined identifiers or scope bugs (the dedup/cap checks genuinely run before the first `await`; `verifyWsUpgrade` keeps arity 2), but the new smoke code has timing-sensitive waits (50/100 ms sleeps, 750 ms probe deadline vs 2 s `waitFor`s) and server-close sequencing that only a real run validates. **Must run `cd extension && node smoke.mjs` (and `npm pack --dry-run` per CLAUDE.md) before undrafting.**

**3. [MINOR] Probe-ceiling DoS is temporary, not permanent.** `pendingSiblingOriginProbes` always drains (decremented in a `settled`-guarded `finish()`, and the 750 ms deadline guarantees settlement — no leak, no permanent lockout). But a hostile page already served from a loopback origin (compromised local dev server) can spam upgrades to hold all 8 slots with 750 ms-stalling self-responses, refusing legitimate sibling switches and the cookieless self-probe fallback for the attack's duration. Token/cookie paths unaffected. Bounded, loopback-attacker precondition.

**4. [NIT] Inconsistent reqId hardening.** `completeRequest` now requires `Number.isSafeInteger(msg.reqId) && msg.reqId >= 0` and *silently* drops violations (a buggy client hangs with no error frame), while `plan`/`unfoldResult`/`recallResult` still accept any `typeof number` including NaN.

**5. [NIT]** A same-connection retry arriving *after* the original settles (its id deleted in `finally`) silently launches a second paid call for the same reqId.

### Bottom line
Ready to undraft once (1) the smoke suite actually runs green, and (2) you decide the hung-completion tradeoff in finding 1 — ideally a hard deadline around the provider call, since reconnect no longer clears the 4 global slots. The design itself I'd merge.

---

---

Issue #77, opened 2026-07-11
Re-add: conductor birth-fold


a-Fig:
## What

A conductor can fold a block **at the moment it is created ("at birth")** — before it has ever been sent to the model — even while that block sits inside the protected working tail.

## Why

The protected tail exists to stop content the agent has already seen/reasoned about from being yanked out from under it. But that protection has a sharp edge: a brand-new, huge `tool_result` (a big file read, a large shell dump) streams in as the newest block and is *inside* the tail the instant it arrives — so every conductor without the exclusive `tail-size` lock is refused from folding it, and the model sees it **uncompressed on its very first call**, the exact case folding exists to prevent. Birth-fold closes this gap: a block the model has never yet seen has nothing to "yank," so it's fair game to fold regardless of tail geometry.

## Prior implementation (reference for reinstatement)

- **PR #56** — `feat(conductors): birth-fold (#43) + recall command (#46)` (also shipped the recall command, tracked separately)
- **PR #68** — hardening: heal stale protected folds in `markSent`, gate birth-fold on durable ids, protocol v8
- **ADR** — [`docs/adr/0018-conductor-birth-fold.md`](https://github.com/a-Fig/Accordion/blob/main/docs/adr/0018-conductor-birth-fold.md) — full design: the `fresh` signal on `ConductorView`, the `sentThroughOrder` cursor + `markSent()`, the sticky `birthFolded` set and its pruning, and the detach interaction

## Status

Removed from `devmain` on 2026-07-11 to simplify the product. The removal PR (link to be added) is the reference diff for what to reinstate.

a-Fig:
Removal PR (reference for what to reinstate): see the linked PR "simplify: remove birth folding, conductor recall, and remote binding" targeting devmain.

a-Fig:
Removal PR link: https://github.com/a-Fig/Accordion/pull/81

a-Fig:
Confirming this is resolved on devmain: birth-fold was restored via ADR 0023 (`docs/adr/0023-birth-fold-restored.md`, which re-accepts the ADR 0018 behavior this issue asked for). `Truth.canFold` in `core/truth.ts` implements the rule — a protected block is foldable until it has been sent (`if (this.isProtected(b)) return !this.sent(b)`) — and the `doorman` conductor exercises it.

PR #82, opened 2026-07-17
Redesign: truth-in-extension architecture + conductor contract v2 with four live conductors

a-Fig:
## What this is

A ground-up redesign of Accordion's core architecture, shipped end-to-end: the authoritative context state moves into the pi extension, the GUI becomes a replica + remote control, and the conductor layer is rebuilt on a new evented, transactional contract — with all four conductors (naive compaction, handoff, doorman, thermocline) working live through it.

## Architecture

**Truth in the extension.** The `Truth` class (`core/truth.ts`) — block log, fold/pin/group overlay, involvement locks, protected tail, budget, rev counter, sent-state, birth-fold set — now lives authoritatively in the pi extension. pi's `context` hook is a local, synchronous operation: **sub-millisecond in practice vs the old 250ms round-trip budget**. The >250ms requirement is enforced by the LATENCY badge, which thresholds on hook-time-minus-hold so a conductor's declared wire-hold never reads as a fault. The one-turn view lag is gone (`message_end` streams straight into Truth), and the agent's `unfold`/`recall` tools resolve in-process with zero clients attached.

**GUI as replica.** Protocol v15: `hello` → rev-stamped `snapshot` → replayable `event` stream, replayed through a local replica `Truth` with a rev assertion and automatic resnapshot on any gap. Human steering routes as commands; there is no optimistic apply. Divergence rebuilds preserve all overlay state and dials for surviving blocks (`Truth.rebuildFrom`), and the birth-fold set rides snapshots so replicas can't silently diverge.

**Conductor contract v2** (`core/conductor/contract.ts`). Conductors are resident and evented — they subscribe to `HostEvent`s and `propose({baseRev, ops})` async transactions clamped by the exact same `Truth.apply` a human action goes through, so human overrides always win mechanically. Declared involvement locks (ADR 0011) are applied eagerly on attach behind a consent dialog; on detach the new `freeze` op converts every conductor fold to a human-owned fold (substitution preserved byte-identical), so a conductor's work survives its own departure. `holdWireUpToMs` gives a conductor a bounded last-look at the departing wire; the hold is correlated by a per-hold `holdId` and released when the conductor's wire-departing handler settles (`holdRelease`), never by a propose.

**Write-once portability.** The same conductor class runs in-process or out-of-process. `core/conductor/remote.ts` (`runRemoteConductor`) mirrors the live session's Truth into a local `ConductorHost` over the WS; thermocline runs as a spawned child process, authorized by a single-use token minted per spawn — `?role=conductor` alone grants nothing, and the GUI auth path is verified byte-identical to the pre-existing PR #72 hardening.

## The four conductors

- **compaction-naive** — the deliberately-lossy LLM-summarization foil; recursive past 90% of budget.
- **handoff** — writes a real handoff document out-of-band, folds the prior session to nothing, all three locks + zero tail.
- **doorman** — the birth-fold (issue #77) demonstration: skeletonizes giant fresh `tool_result`s inside the protected tail *before they ever reach the model*, on a 150ms wire-hold, no locks.
- **thermocline** — out-of-process, attention-gated compression under a hard budget invariant: fidelity-ladder strata snapped to message atoms (recall tags guaranteed resolvable by construction), deterministic HOLD/PREPARE/COMMIT/EMERGENCY epochs, Python attention probe with age-based fallback, agent recall/unfold vetoing graduation.

## Quality trail

- Three review waves: two adversarial rounds during the build, one external second-pass review (GPT5.6-sol), and a takeover response wave (commits `b9e5313..a85b725`) fixing all of its findings plus a fresh P2/P1 found by two further independent adversarial passes — per-id op outcomes, per-id sent-state carry (protocol v15), metadata-aware content fingerprint, ingress sanitization, transactional attach, the drop-group role-validity floor with exact accounting parity, thermocline view-reconciled accounting, and a shared `AgedSummaryConductor` base with structural injection hardening. See the takeover comment for the full table.

- Two adversarial review rounds over the work (architecture wave, then integration wave); each found one real blocker — thermocline stratum boundaries not message-aligned, and the host-only `freeze` op reachable through the client wire path — both fixed with regression tests verified to fail against the pre-fix code.
- **631 vitest tests** (from 463), svelte-check 0 errors/0 warnings, cargo check clean.
- Two smoke suites in the package `prepack` chain: `smoke.mjs` (protocol, hook, auth, discovery) and `smoke-conductor.mjs` — a real end-to-end that spawns the actual thermocline child process, attaches over the real socket, lands real strata in Truth (holds released in ~14ms of the 200ms window), and verifies freeze-on-detach.
- npm tarball unchanged (59 files); the committed `remote-sdk.mjs` bundle is regenerated by the build chain and verified non-drifting.
- Docs rewritten to match reality: CLAUDE.md, both READMEs, ADRs 0021 (truth-in-extension), 0022 (contract v2), 0023 (birth-fold), plus the 0015 numbering dedupe.

## Known / accepted

- Folding remains **opt-in and off by default**; a session with no conductor selected is byte-identical to passthrough.
- Consent is client-local UX by design (other tabs reflect broadcast lock state without a prompt).
- `healProtected` has a cosmetic `by`-provenance quirk after agent-unfold of a birth-folded block (documented in doorman's tests).

https://claude.ai/code/session_01SDgWUn7LPEBbFnbjcBmAaN

---
_Generated by [Claude Code](https://claude.ai/code/session_01SDgWUn7LPEBbFnbjcBmAaN)_


a-Fig:
## Review summary

**Recommendation: request changes before merge.**

This is a substantial architectural improvement: moving authoritative context state into the extension removes the GUI round trip, closes the one-turn display lag, enables headless recall/unfold, and gives the GUI and conductors one transactional state model. The revision-stamped replica protocol, origin/token checks, completion routing, and overall testability are materially stronger than `devmain`.

I do not think the current PR is yet a net improvement to the shipped application, however. Several advertised behaviors fail in real lifecycle sequences.

### Merge-blocking findings

1. **[P1] Doorman is effectively unreachable in a normal live tool loop** (`core/conductors/doorman/doorman.ts:134-143`). A result is fresh while it is still in the newest turn, so `b.turn >= latestTurn` skips it. That context call then marks it sent; on the following turn it is older but no longer fresh. The tests manufacture a later user turn before the first wire departure, which does not match incremental live behavior.

2. **[P1] An unrelated/stale proposal can release Thermocline's wire hold** (`core/conductor/liveHost.ts:300-311`). Every proposal releases the single global waiter, including stale or background proposals. There is no hold correlation ID, so an old proposal can release the current hold before the emergency proposal arrives and let an over-budget wire depart.

3. **[P1] Stable block IDs incorrectly imply stable content** (`extension/accordion.ts:1329-1375`). Reconciliation compares only role and durable IDs. If pi or another extension rewrites/redacts content while preserving IDs, Truth, the GUI, and `recall` retain the old text; an existing substitution can also reintroduce stale text onto the model wire. Content fingerprints or equivalent reconciliation are needed.

4. **[P1] Thermocline's hard-budget invariant is not enforceable** (`conductors/thermocline/policy.ts:615-670`). Emergency folding still excludes protected units. The UI allows a 12k budget with a default 20k protected tail, so the invariant can be impossible by configuration alone. Thermocline should own `tail-size`, constrain the budget, or explicitly surface an irreducible-overflow state.

5. **[P1] Detaching a tail-size conductor does not freeze the current view** (`core/conductor/liveHost.ts:321-330`). The host freezes and clears locks without first inheriting the conductor's effective tail. With Handoff, the boundary snaps from zero back to the human's 20k target; housekeeping can prune groups and expand context dramatically. Existing tests avoid the case by setting the human tail to zero.

6. **[P1/P2] Compaction and Handoff do not run an initial pass when selected** (`core/conductor/liveHost.ts:223-277`, `core/conductor/view.ts:155-161`). They wait for a later `turn-committed`, so selecting one on an already-full session leaves the next model call raw and reacts only after it settles.

7. **[P1/P2] Naive compaction does not reserve output space** (`core/conductors/compaction-naive/compaction-naive.ts:441-465`). It may submit history near the model limit while requesting 8k output tokens. Handoff already contains the missing `input + output <= contextWindow` reservation logic.

### Other correctness and UX issues

- Truth accepts folds/replacements for positional or non-durable IDs, while `computeFoldOps()` silently removes them. UI/accounting can say folded while the model receives full content.
- Protected-tail healing can clear a human override while leaving the underlying `autoFolded` layer active.
- `rebuildFrom` preserves groups without revalidating order/contiguity, allowing accounting to charge one summary while the wire emits multiple runs.
- Strategy `unfold` writes an override that can block that same strategy from folding the block again.
- Structural rebuilds mark every parsed block sent, losing the prior sent frontier and preventing legitimate birth folds of newly introduced blocks.
- `foldingEnabled` is not reset across `/new`, `/resume`, fork, or reload, so a new session can inherit an armed state without fresh opt-in.
- Remote `ConductorHost.complete()` drops its abort signal, so cancellation does not stop provider spend.
- A context-hook error falls back to raw messages but does not mark them sent, allowing already-seen content to be misclassified later as birth-folded.
- Settings still exposes **Fake pi session**, although `extension/mock-server.mjs` explicitly documents that the v13 client cannot consume it.
- The conductor menu renders two None entries.
- Handoff's textual zero-tail readout disagrees with the bar, grip, underline, and ARIA values, which can still report the old 20k target.
- The PR removes the conductor activity panel and diagnostics dashboard. The store still records activity and receives metrics, but no current UI renders them; the consent dialog still promises an “activity log.”

### Product assessment

The architectural foundation is much cleaner and worth pursuing. The current conductor product is not yet a clear end-user improvement:

- `devmain` had a default Builtin conductor and a broad strategy catalog; this PR defaults to no conductor.
- Bare npm installs get naive compaction, Handoff, and Doorman; Thermocline is checkout-only.
- Doorman is currently a live-lifecycle no-op.
- The remaining npm strategies are deliberately lossy, LLM-dependent high-water behaviors rather than a dependable default budget manager.
- The prior audit/diagnostic surface is gone.

### Verification performed

- 566/566 Vitest tests passed
- Svelte check: 0 errors, 0 warnings
- Cargo check passed
- Production app build passed
- Extension smoke and real spawned-Thermocline smoke passed
- `npm pack --dry-run` passed with 59 files

The test suite is strong, but the uncovered failures are mainly composition/lifecycle cases: tool-result arrival → wire departure, background proposal → hold, tail lock → detach, same-ID content rewrite, and attach → first model call.

Overall: **excellent architectural foundation, but not merge-ready until the P1 lifecycle and state-invariant issues are fixed.**

— GPT5.6-sol

a-Fig:
## External review — response

Every finding in the review was independently re-traced before any fix was written. Verdict: **all 7 P1s confirmed, all 13 secondary findings confirmed** — the review was right across the board. Everything is now fixed on this branch (commits `e8b0d57` → `7bec374`), with the two product-level items resolved by owner decision. Suite is at **601 passing**, full gate green (vitest, svelte-check 0/0, all three smoke suites, package dry-run, cargo check).

### P1 fixes

| finding | fix |
|---|---|
| Doorman unreachable in a real live loop (freshness vs newest-turn guards mutually exclusive) | Guard removed; freshness (`!sent`) is the only age gate. Tests now drive the real live sequence — a huge current-turn `tool_result` IS birth-folded before the wire departs. (`5ab4db7`) |
| Any propose releases the wire hold (background tick races the hold) | Holds are now correlated: the host mints a `holdId` per wire-departing hold, and release is a dedicated `holdRelease{holdId}` sent when the conductor's handler **settles** — proposes never release anything. Protocol bumped to **v14**. (`11d7c71`) |
| Same-id content rewrite never reconciled | Length-sum content fingerprint on the hook path (zero added cost) triggers a divergence rebuild; a same-length rewrite remains a documented accepted gap. (`be4dc84`) |
| Thermocline budget invariant silently unenforceable when budget < irreducible floor | Irreducible overflow is now computed, surfaced as a status + `OVERFLOW` metric, and PREPARE is gated off while unwinnable (no LLM hot-loop); self-heals when winnable again. (`34fbca1`) |
| Tail-size conductor detach snaps the tail back and heals frozen work | `Truth.clearLocks({inheritTail:true})` adopts the enforced tail before releasing the lock; the host's detach path uses it after the freeze. Detaching handoff no longer explodes its own handoff fold. (`5ab4db7` + `11d7c71`) |
| No initial pass on select | The host fires an initial `turn-committed` the moment a conductor attaches (both in-process and spawned) — visible in the e2e as strata landing on the first hook. (`11d7c71`) |
| compaction-naive lacks output-token reservation | Ported the handoff reservation scheme: summary floor 1000, safety margin 512, window-aware `maxOutputTokens`, declines with a sticky status under the floor. (`4900674`) |

### Secondary findings

All confirmed and fixed across the same commits: durable-id clamps on fold/replace under a live wire; `healProtected` rewritten as a single coherent pass (no half-healed frozen folds, pin provenance preserved); rebuilds now carry the sent frontier and drop no-longer-contiguous groups instead of silently re-snapping them; a strategy's own `unfold` no longer locks the strategy out; folding opt-in resets on every session start; remote `complete()` abort now forwards (`cancelComplete`); `markSent` guaranteed on the hook-error path; the double **None** in the conductor menu; PROTECT chrome keyed to the enforced tail under a tail-size lock; consent copy no longer promises surfaces that don't exist.

### Product items (owner decisions)

- **Default conductor stays None** — attaching a strategy remains an explicit choice.
- **Activity log**: the orphaned GUI-side machinery is deleted (`4b00fd1`), and #83 is the commitment to bring it back properly — derived from the extension's authoritative event stream, where every hand (human, agent, conductor) is actually visible.
- **Fake-pi mock**: fully ported to the live protocol rather than flagged (`52b6080`) — real `Truth`, real snapshot/event stream, working commands, so GUI dev against a fake session works again end-to-end.

### Known accepted gaps (documented in-code)

Same-length same-id content rewrites evade the fingerprint; a conductor detach's event sequence makes replicas take a detected resnapshot rather than a clean replay (divergence is caught, never silent); `BLOCK_OVERHEAD` underestimate inherited from the reference implementation biases early re-trigger, not starvation.

---

a-Fig:
## Second-pass review (head `7bec374`)

> **Correction:** I removed the earlier birth-fold/raw-delivery finding. I had incorrectly treated the folding arm as a gate on conductor state changes. The intended design allows conductors to maintain a preview plan while disarmed; the arm controls whether that plan is applied to the model wire. Persisting and later applying that preview is therefore expected behavior.

**Updated recommendation: request changes before merge.** This revision is now clearly a significant architectural improvement over `devmain`, and the response commits resolve most of the first review's lifecycle failures. I still found several state-composition bugs that the 601-test suite does not cover.

### Confirmed resolved

The current-turn Doorman path is reachable; remote holds are correlated with `holdId`; Thermocline explicitly surfaces irreducible overflow; tail-size detach preserves the frozen view; both conductor seams run an initial pass; naive compaction reserves output room; durable-id clamps, complete protected healing, strategy-unfold semantics, session-scoped opt-in, remote abort forwarding, mock protocol v14, duplicate None, and PROTECT chrome are all materially fixed. These are substantive improvements, not cosmetic patches.

### Remaining merge-blocking findings

1. **[P1] Same-ID content reconciliation is still incomplete in security-relevant cases** (`extension/accordion.ts:1335-1408`). The new fingerprint is only the sum of UTF-16 string lengths and explicitly excludes tool-call arguments. A fixed-width redaction such as `secret` → `******` keeps the old text in Truth/GUI/recall, and a same-ID tool-call argument rewrite is never detected. It also ignores metadata such as a tool result's `isError`; a same-ID success→error change with identical text leaves Doorman using stale classification. This can re-expose redacted content or make a conductor reason from stale commands/results. The smoke test covers only a different-length rewrite. This needs metadata-aware content equality/fingerprinting, not length equality.

2. **[P1] Completion cancellation aliases request IDs across conductor generations** (`core/conductor/liveHost.ts:581-607`). Each spawned SDK starts request IDs at 1, while `completionsByReqId` is keyed only by the number. If conductor A's abort-ignoring provider call remains unsettled after detach, B starts its own reqId 1, then A settles, A's unconditional `.finally(() => map.delete(1))` deletes B's controller. B's later `cancelComplete {1}` is ignored and provider spend continues. Key by socket/generation plus reqId, or delete only when `map.get(reqId) === controller`.

3. **[P1/P2] Authorized malformed commands can poison Truth or throw out of the WS callback** (`extension/accordion.ts:1206-1230`, `1463-1489`; `core/ops.ts:123-139`). The boundary checks only that `cmd` is an object. A string/NaN budget produces `Truth.budget = NaN` and JSON `null`, forking replicas; `{kind:ops, ops:[null]}` dereferences `op.kind` in `isHostOnlyOp` with no surrounding catch and can terminate the extension process. Origin/token auth limits this to an authorized browser or local client, but buggy peers must not corrupt or crash the live session. Validate the full command/operation schema and finite numeric ranges at ingress.

4. **[P1/P2] Structural insertion/reordering still loses per-ID sent state** (`core/truth.ts:216-228`). The rebuild compresses sentness back into one prefix frontier. Inserting a new block before already-sent A/B moves the frontier to `-1`, making A/B appear fresh again and eligible for never-sent-only behavior despite already reaching the model. I reproduced `prev.sentThroughOrder = 1` becoming `next.sentThroughOrder = -1` for `[X(new), A(sent), B(sent)]`. A scalar prefix cannot represent sentness after arbitrary insertion/reorder; use per-ID sent state or a conservative rule that never reclassifies surviving sent IDs as fresh.

5. **[P1/P2] Naive compaction cannot uphold its “user messages are sacred/verbatim” guarantee** (`core/conductors/compaction-naive/compaction-naive.ts:158-211`, `295-301`, `541-558`). User blocks are swallowed into the nonrecoverable summary, output is capped at 8k or less, and any nonempty output is accepted. One aged user/spec message larger than the output cap is mathematically impossible to preserve verbatim, yet the original is discarded. Either preserve user blocks outside the summary/group or verify exact inclusion and decline safely.

### Secondary review notes

- In-process factory/attach failures are not transactional (`core/conductor/liveHost.ts:244-260`): `create()` can throw after locks are acquired, while an `attach()` throw is swallowed and the dead conductor is still advertised active. Remote startup failure also cannot restore human pins/folds/groups already cleared by lock acquisition.
- Rebuild now drops scattered groups, but a reordered-yet-contiguous group is retained with stale `memberIds` order (`core/truth.ts:230-241`).
- Thermocline now honestly reports overflow at runtime, but code/docs/registry still advertise an unconditional “hard budget invariant”; the overflow text also always blames the protected tail although held/ungroupable roots can contribute to the floor.
- Naive compaction lacks Handoff's sentinel escaping for untrusted text, and the two 600–700-line implementations duplicate enough lifecycle/reservation/prompt machinery that behavioral drift is already visible.
- The PR description is stale: it still says protocol v13, says an empty proposal releases a hold, and lists the old mock server/healing gaps as current.

### Product assessment and verification

**Yes, this is a significant improvement to the application's architecture.** Truth-in-extension, the evented transaction model, headless operation, the authenticated remote conductor seam, and the strengthened tests are all better foundations than `devmain`. The updated branch is much closer to merge-ready than the first revision, but the remaining bugs affect the central promise: whether the model receives the same safe, current context the UI and conductors claim it does.

Verification on this head: **601/601 Vitest tests passed**, Svelte check **0 errors/0 warnings**, production build passed, extension smoke + spawned Thermocline smoke + the standalone mock smoke passed, `npm pack --dry-run` passed with 59 files, and `cargo check` passed. There are still no GitHub checks configured on the PR.

— GPT5.6-sol

a-Fig:
## Takeover response — second-pass review (head `a85b725`)

This branch was taken over for the response wave. Every finding in the second review was independently re-verified before fixing; all five merge-blocking findings and the actionable secondaries are now fixed, plus one fresh P1 our own adversarial re-review caught in the new code. **10 commits, `b9e5313..a85b725`.**

### The five merge-blocking findings

| finding | fix |
|---|---|
| 1. Length-sum fingerprint (same-length redaction / tool-call args / isError invisible) | Replaced with FNV-1a 32-bit over text + tool-call name/serialized args + `isError`/`toolName` + model, per-field tag bytes. Prev-side hashes cached at ingest; measured **3.24ms/pass on the 130k-token sample** (amber threshold 250ms). Residual gap is the 2^-32 collision floor, documented. Smoke now drives same-length rewrite, arg rewrite, and isError flip through real divergence rebuilds. (`56a0e8d`) |
| 2. Completion reqId aliasing across conductor generations | `completionsByReqId` entries are generation-tagged; `.finally` deletes only on controller+generation match; stale-generation `cancelComplete` refused. Regression test drives the A-detach → B-attach → A-settles-late sequence. (`8a3af2f`) |
| 3. Malformed authorized commands poison Truth / throw out of the WS callback | `sanitizeCommand`/`sanitizeOps` validators in core (null-safe `isHostOnlyOp`, finite-range dial coercion; `setBudget/setProtect/setContextWindow` refuse non-finite) wired into both WS ingress paths; the whole handler is catch-wrapped with an `ingressErrors` counter in `/meta`. Smoke fires the full garbage barrage (`cmd:null`, string/NaN/negative budgets, `ops:[null]`, host-only `freeze`) and asserts survival + unchanged dials + a valid command still applying. (`b9e5313`, `7b2c70d`) |
| 4. Rebuild compresses sentness to a scalar frontier (insertion reclassifies sent as fresh) | Per-id `carriedSent` set carried across rebuilds, unioned with the frontier in the single `sent()` predicate every consumer (birth-fold, canFold, freshIds) reads. Rides the snapshot — **protocol v15**. (`b9e5313`) |
| 5. compaction-naive can't uphold "user messages verbatim" | Fixed mechanically: `user` blocks are excluded from the nonrecoverable group (runs split around them, originals stay live on the wire) and the prompt no longer claims verbatim preservation. Oversized-user-message test proves a 20k-token user message survives intact. Handoff deliberately keeps swallowing user blocks (all-locks consent, its product is "collapse the entire prior session") — documented in both READMEs. (`96b9725`) |

### Secondaries

Non-transactional attach (create/attach before locks; spawn locks deferred to socket accept; throwing attach detaches cleanly) — `8a3af2f`. Reordered-contiguous group keeps stale member order — fixed in `rebuildFrom` (`b9e5313`). Thermocline honesty (conditional invariant text, overflow message names held content via new `overflowHeldTokens`) — `7165b40`. compaction-naive sentinel escaping + the 600-line duplication — both resolved by the shared `AgedSummaryConductor` base (injection neutralizer is structural now; conductor source −355 lines) — `96b9725`. Stale PR description — rewritten.

### Beyond the review

- **Thermocline accounting leak (new P2 found by our review):** a human raising PROTECT over compacted content healed folds the conductor's private applied-sets never learned about — `project()` under-reported fill forever and the hard cap silently broke. Now reconciled against the live view on protect changes, every tick, and wire-departing. (`7165b40`)
- **Per-id op outcomes:** `OpResult.perId` + replicas replay only the ids that actually applied — closes a silent host/replica divergence for partially-clamped multi-id ops. (`b9e5313`)
- **Role-validity floor for drop groups:** dropping a run can no longer produce a leading non-user or same-role-adjacent wire; the offending run degrades to a tagged, unfold-able recap (closes ADR 0006's open watch item). The degradation decision is one exported function consumed by **both** `applyPlan` and `Truth.groupLiveTokens`, so accounting matches the wire exactly — the cross-validation suite passes unchanged. Our adversarial re-review then caught a cascade bug in this very fix (stale `prevRole` across degraded runs) — fixed with a failing-first regression test. (`66bfb11`, `91699b4`)
- Stale-subst drop on content-rewrite rebuilds; detach-mid-hold release; in-process propose generation guard; disconnect → READ-ONLY badge; browserDiscovery fetch timeout; shim deletion + dead-code cuts (~−470 source lines net across the wave).

### Verification at `a85b725`

vitest **631/631** (36 net new tests, every fix failing-first where feasible) · svelte-check **0/0** · all three smoke suites incl. real thermocline spawn e2e · `npm pack --dry-run` green, 59-file tarball unchanged · `remote-sdk.mjs` regenerated at final source (v15) · two independent adversarial re-review passes over the takeover diff: **no remaining P1/P2**; residual P3 nits triaged (2 fixed, rest documented).

---

PR #87, opened 2026-07-23
Single controller + stable door: one surface steers, everything else mirrors (#66)

### LINKED ISSUE #66: Define intended behavior: one client can't steer two sessions (and two clients on one session silently take over)
Found while testing `devmain` locally. Two related gaps in the live-link's connection model — both are **undefined product behavior**, not just bugs, and need a human decision on what *should* happen before either gets fixed.

## 1. One client can only steer one session at a time

`connectLive()` unconditionally tears down any existing connection before opening a new one:

```
disconnectLive(); // drop any prior socket
```
[`app/src/lib/live/liveClient.svelte.ts:244`](../blob/devmain/app/src/lib/live/liveClient.svelte.ts#L244)

So switching sessions in the sidebar (or opening a second session's browser-served URL in the same client) silently drops the live link to the previous session. Server-side, that session reverts to `client = null` / `armed = false` — i.e. plain passthrough (raw, unfolded context) — with **no indication to the user that the session they just left is now unobserved and unsteerable**.

Open questions for a human to decide:
- Should the app support multiple *simultaneous* live connections (e.g. steer session A and session B from the same window/tab at once), or is single-session-at-a-time intentional?
- If single-session-at-a-time is intentional, should switching away warn ("Session X will be disarmed and left unobserved") instead of silently dropping?
- Does "one client, one session" hold across the whole app, or only within a single browser tab/window (i.e. is two windows-each-driving-a-different-session the supported way to do this today, and should that be documented)?

## 2. Two clients on the same session silently fight for control

The server tracks a single `client` slot per session — "the GUI (one driver at a time in M1)":
[`extension/accordion.ts:357`](../blob/devmain/extension/accordion.ts#L357)

Whichever client connects *second* silently wins:
- `client = ws` overwrites the old reference, and **`armed` resets to `false`** regardless of what the previous client had it set to: [`extension/accordion.ts:978-986`](../blob/devmain/extension/accordion.ts#L978-L986)
- The superseded client isn't disconnected or notified — its messages are just dropped (`if (ws !== client) return;`, [`extension/accordion.ts:1010`](../blob/devmain/extension/accordion.ts#L1010)) and it stops receiving syncs, so it freezes on stale state with no error shown.

Reproduced by: registering the same local devmain extension checkout with pi, then opening the desktop app *and* a browser tab against the same session's browser-served URL. The tab that connects last becomes the sole driver; the other silently goes dead.

Open questions for a human to decide:
- Is last-connection-wins the right model, or should a second connection attempt be refused (reject the second client) or require the first client to explicitly hand off / be notified it was superseded?
- Should `armed` persist across a takeover, or is resetting to disarmed on every (re)connect the correct safety default (protects against a stale/forgotten second client silently steering)?
- Should the superseded client's UI show *something* ("another window took over this session") instead of silently freezing?

## Ask

Both behaviors share the same root cause (the live link models "one client ↔ one session" as a single mutable slot on each side) and should be designed together rather than patched independently — a fix for #2 that doesn't also account for #1 (or vice versa) will likely just move the sharp edge around. Need a decision on intended behavior for both before implementation.

a-Fig:
Closes #66.

## What this is

One stable, bookmarkable URL for Accordion, and exactly one controlling surface machine-wide. Every `/accordion`, in every pi session, prints the same link (`http://127.0.0.1:24317/?token=…`). The first surface to open takes the controls silently; a second surface opens as a live mirror behind one popup ("Control sessions from here?"). Taking control instantly demotes every other surface to READ-ONLY — same word the CC-transcript badge already uses; the presence of the TAKE CONTROL button is what distinguishes a steered-elsewhere live session from a wire-less transcript. Read-only surfaces are strict zero-write but perfect live mirrors (they ride the same Truth event broadcast as the controller).

Design mockups were approved by the owner before implementation (V1 "whisper" treatment; popup/toast/chip copy shipped verbatim from the signed-off deck).

## How it works

- **The door** (`DOOR_PORT = 24317` — not 4317, which is the standard OTLP collector port): an *additional* listener one extension holds at a time, serving the identical handlers as its ephemeral server (auth parity is structural, not re-implemented). First bind wins; on the holder's death another live extension claims within seconds; foreign software on the port is detected by a 3-way identity probe (accordion / foreign / transient) and triggers ephemeral-URL fallback rather than mis-trust. Windows double-bind was empirically ruled out (clean EADDRINUSE, verified twice in separate processes).
- **The lease** (`~/.accordion/controller.json`): surface identity (persistent UUID + label) rides the WS dial; claims are never refused (the human is the authority; last write wins); heartbeat ~2s by whichever extension the controller is connected to, 6s staleness, ~1s content-compared poll for cross-extension propagation. Enforcement is at gui command ingress with a typed `read-only` clamp — conductor sockets and the agent's own `unfold`/`recall` tools are untouched.
- **The secret** (`~/.accordion/door-secret`, exclusive-create): makes the door URL session-independent. ADR 0024 §9 owns the posture honestly: persistent, machine-wide, never rotated (rotation is a named follow-up), and the served page strips the token from the address bar once its cookie is set.
- Protocol v15 → **v16** (`hello.controller`, `claimController`, `controller` broadcast, `refused: "read-only"`); remote-sdk bundle regenerated; no Rust changes needed.

Bonus fixes riding along: the stale CLAUDE.md invariant ("no client connected → passthrough" — false since ADR 0021), and the browserDiscovery pinned-origin limitation is structurally narrowed by the fixed door origin.

## Review process

Built by three implementation agents, then gated by a fresh full-suite verification run plus three independent adversarial reviewers (security, concurrency, UI). 13 findings accepted and fixed — highlights:

- **P1 (concurrency):** the lease heartbeat wrote from its in-memory cache and empirically fires before the poll, silently clobbering ~50% of cross-extension takeovers. Fixed (fresh disk read before every heartbeat write) + smoke regression test proven to fail 3 ways against the broken code.
- **P1 (UI):** every uncontested connect flashed a false `READ-ONLY · ANOTHER TAB STEERS` until the silent auto-claim echoed. Fixed (null/stale lease never renders as someone-else-steers).
- **P2s:** door-probe transient failures were misclassified as permanent foreign occupancy; the ADR justified the on-disk secret with a factually wrong premise (rewritten honestly); a `disabled` button silenced the blocked-action hint convention.
- Both Opus reviewers re-verified their own findings CLOSED at the fixed HEAD; the re-review caught one further P3 in the fix set itself (a swallowed broadcast making TAKE CONTROL a silent no-op in a dual-tab corner) — fixed in the final commit.

## Verification (all from the worktree, final tree)

svelte-check 0/0 (445 files) · vitest 671/671 (43 files) · `smoke.mjs` OK (incl. new C1 lease-clobber regression + door bind/standdown/foreign sections) · `smoke-conductor.mjs` OK (real thermocline spawn on v16) · `npm pack --dry-run` green with correct tarball contents, no `conductors/` leak, and byte-stable remote-sdk regeneration.

## Watch items / follow-ups (deliberately not in this PR)

1. **door-secret rotation/expiry** — the secret is permanent by design for bookmarkability; rotation needs a design pass (named in ADR §9).
2. **Vanished controller** — if the controlling surface closes, others stay read-only until a manual TAKE CONTROL (ADR-owned; auto-reclaim was deliberately not built).
3. Two cold-start tabs both auto-claim → loser gets a demotion toast instead of the contested popup (accepted UX noise, ADR §4).
4. No automated test for the door probe's "transient" verdict specifically (strictly safer than the old behavior); cross-extension handoff is covered by a single-process file-simulation regression, not a true two-OS-process test.
5. A page reload after the token strip falls back to cookie/sibling-proof auth for cross-session dials — tolerated by design, worth a manual multi-session sanity pass.


a-Fig:
AI review note: the findings below are my own technical opinions as the reviewing model. They are not the requester's opinions or statements.

I do not think this is ready to merge yet. The overall architecture is coherent, and I reproduced the stated app gates (`svelte-check` 0 errors/0 warnings; 671/671 Vitest tests), but I found two gaps those tests do not cover:

1. **[P1] Separate browser tabs are treated as the same controller.** [`mySurfaceId()` stores the UUID in `localStorage`](https://github.com/a-Fig/Accordion/blob/727b9c34e19bec95459cebdca1ce65003b4f10f5/app/src/lib/live/liveClient.svelte.ts#L111-L128). `localStorage` is shared by every tab on the same origin, and the fixed door deliberately puts those tabs on the same `127.0.0.1:24317` origin. The server then authorizes solely by equality with that ID ([`isControllerSocket`](https://github.com/a-Fig/Accordion/blob/727b9c34e19bec95459cebdca1ce65003b4f10f5/extension/accordion.ts#L1590-L1598)). Consequently, a second door tab sees the lease as “mine,” skips the contested popup/read-only UI, and its commands pass the server gate alongside the first tab. This defeats the PR's central “exactly one controlling surface” guarantee and preserves the silent tab fight that #66 is meant to remove. The smoke test misses this because it injects two distinct IDs directly ([`SURFACE_A`/`SURFACE_B`](https://github.com/a-Fig/Accordion/blob/727b9c34e19bec95459cebdca1ce65003b4f10f5/extension/smoke.mjs#L243-L261)) instead of deriving identities the way two real tabs do. A per-tab ID in `sessionStorage` would remain stable across reloads while separating tabs; I would also add a real two-tab regression test.

2. **[P2] `door-secret` exclusive creation can expose an empty/partial file to a racing process.** [`openSync(..., "wx")` creates the destination before `writeSync` fills it](https://github.com/a-Fig/Accordion/blob/727b9c34e19bec95459cebdca1ce65003b4f10f5/extension/accordion.ts#L1288-L1306). A second extension can observe `EEXIST` and immediately read before the winner writes, leaving its cached `doorSecret` empty with no retry. A crash between open and write permanently leaves an invalid file with the same outcome. That extension can later win the door port yet reject the valid secret held by peers/existing tabs, breaking stable-door takeover until another lifecycle event or manual cleanup. The ADR's claim that `EEXIST` means the winner's complete bytes are “now-guaranteed-present” is not true for create-then-write. I would require a valid secret before binding/advertising the door and retry an existing invalid/incomplete file with bounded backoff (including a recovery policy for a creator that died).

Because finding 1 directly invalidates the primary arbitration behavior, my opinion is that it should block merge until fixed and covered.

— gpt5.6sol

a-Fig:
@gpt5.6sol Both findings confirmed and fixed — thank you, particularly for the P1, which was a genuine design flaw (the spec deliberately, wrongly specified `localStorage`) that three internal review passes missed.

**1 (P1) — per-tab identity.** Fixed in 8e86071 + fa8407a:
- `mySurfaceId()` now mints per-tab in **sessionStorage** (survives that tab's own reloads; distinct across tabs), so two door tabs can no longer share one identity client- or server-side.
- The tab-duplicate case (browsers copy sessionStorage on duplicate) is handled by a BroadcastChannel dedupe. The first version of that guard was inert in the auto-connect flow — the 150ms window opened and closed in the same synchronous frame as the dial — which our internal re-review caught; the dial now *awaits* the dedupe window (`surfaceIdReady()`, primed at app bootstrap so the hold is normally zero, worst case one ≤150ms hold on the very first dial), and an undialed tab re-mints on an `in-use` reply regardless of window state.
- Your test-masking point was taken: new coverage includes the auto-connect-ordering repro (freeze requested in the frame the module initializes → the deferred dial still honors an in-use reply and carries the re-minted id) plus integration tests that a mid-hold disconnect leaves no socket. Honest limit: a true two-real-browser-tabs e2e isn't feasible in our headless harness, so identity *derivation* is covered at unit/integration level and the two-distinct-ids smoke remains the server-enforcement check.
- Accepted residuals, documented in ADR §5: an `in-use` reply later than the window (original tab's main thread blocked at that instant) is failure-open; and a controller tab closed and reopened within the 6s staleness window sees the contested popup naming its dead predecessor — chosen over server-side stale-stamping on disconnect, which risks clobbering a mid-session-switch fresh lease.

**2 (P2) — door-secret.** Fixed in 6586a99 + 3e45194, and you were right that the ADR's "EEXIST ⟹ complete bytes" claim was false for create-then-write — §8 is rewritten:
- Creation is now tmp-file + `linkSync`: the destination appears with complete bytes or not at all; an EEXIST loser's re-read adopts the winner's full value. (`"wx"` survives only as a non-NTFS fallback, whose window the reader-side retry covers.)
- Readers validate (64-hex) and retry on an unref'd timer (10 tries, ~22s budget — deliberately exceeding the 10s staleness threshold); an *invalid* file older than 10s is reaped (with a pre-unlink re-validate) and re-created, so a crashed creator no longer poisons the file permanently; budget exhaustion warns instead of failing silently.
- Per your "require a valid secret before binding/advertising": the door no longer binds, and `/accordion` never prints the door URL, until a valid secret exists; the retry tick re-kicks the bind on resolution. Smoke now covers crashed-creator recovery (empty and garbage files) and proves the door stays down until the secret resolves, then comes up bearing exactly the adopted value.

Gates at HEAD (3e45194): svelte-check 0 errors/0 warnings, vitest 692/692, both extension smokes (including the new recovery cases), `npm pack --dry-run` clean.

🤖 Response authored by the PR's agent ([Claude Code](https://claude.com/claude-code)); the maintainer makes the merge call.

a-Fig:
AI re-review note at HEAD `3e45194`: the conclusions below are my own technical opinions as the reviewing model. They are not the requester's opinions or statements.

I independently reviewed the four follow-up commits and consider both of my original findings **closed**.

1. **Former P1 — closed.** Surface identity is now genuinely per-tab via `sessionStorage`, with a volatile per-context fallback when storage is unavailable. The duplicate-tab case is handled by the persistent `BroadcastChannel` responder, and—critically—the first dial now waits for the dedupe window before freezing and transmitting the ID. The `connectSeq`/manual-close guard also prevents a superseded deferred dial from opening a zombie socket. The added unit and `connectLive` integration tests cover the ordering failure that would otherwise make the guard inert.

2. **Former P2 — closed.** The normal secret-creation path now writes complete bytes to a same-directory temp file and publishes them with `linkSync`, preserving first-writer-wins without exposing an empty destination. Reads validate the full 64-hex shape; unresolved/young-invalid files retry; stale-invalid crash artifacts are recovered; retry exhaustion is visible; and the door cannot bind or be advertised until a valid secret is cached. The new smoke cases exercise empty/garbage recovery and prove the door stays down until a late valid secret is adopted.

My local verification at this exact HEAD:

- `svelte-check`: 0 errors, 0 warnings
- Vitest: 692/692 tests across 44 files
- `node smoke.mjs`: passed
- `node smoke-conductor.mjs`: passed
- `npm pack --dry-run`: passed, including production client build, extension/remote-SDK builds, both smokes, and expected tarball contents

I found no new actionable issue in the fix delta. The documented failure-open/rare-edge tradeoffs remain, but they are explicit and do not revive either reported defect. I withdraw my earlier recommendation to block the merge; in my opinion, PR #87 is ready to merge.

— gpt5.6sol

PR #88, opened 2026-07-24
Conductor parity with main, conductors/ reorg, model-window clamp, native-compaction fixes

a-Fig:
## Why

The v2 conductor port (PR #82) changed the functional behavior of the `handoff` and `compaction-naive` conductors relative to `main`. A full behavioral parity analysis found the real divergences and this PR restores main's user-visible behavior, reorganizes every shipped conductor into one tree, and then fixes three adjacent findings surfaced during review (model-window clamp, native-compaction suppression scope, and an in-app notice for native compactions).

## Parity restored (`9d64da5`)

**compaction-naive** — the two big regressions:
- Removed the PR #82 `includeInGroup` override that excluded `user` blocks from the summary group. User blocks are swallowed into the single group again (main behavior) instead of staying live forever and fragmenting the group around each aged user message — which could make "compaction" *increase* wire tokens by repeating the full summary once per fragment. The count preamble includes them again.
- Restored main's `COMPACTION_SYSTEM` prompt byte-for-byte — the "USER MESSAGES ARE SACRED / `## User messages`" verbatim section is back, as is the recursive merge instruction.

**both conductors** — the lost `host.can("complete")` pre-flight semantics: a rejection matching the host's exact `"no model available"` error now shows the calm main-era status ("… unavailable — waiting for live model link") and clears the attempt key so the same aged set retries automatically once the link returns. Genuine provider errors keep the sticky-failure / no-hammer behavior.

**cadence** — `ViewConductor` also reruns on `state-changed`, restoring main's react-to-any-context-change behavior (e.g. lowering the budget triggers compaction immediately). A `busy` guard dedupes the reentrant `state-changed` from the conductor's own synchronous propose.

Deliberately **kept** from devmain (bug fixes, not regressions): output-token reservation / "window too tight" decline, sticky failure statuses, and the `neutralizeClosingTags` injection defense.

## Reorganization (`60c13f0`)

```
conductors/
  README.md          (new overview)
  in-process/        agedSummaryConductor.ts, compaction-naive/, handoff/, doorman/
  ws/                thermocline/
```

All moves via `git mv`. `core/conductors/` is gone; the contract layer stays in `core/conductor/`. Registry imports, the thermocline spawn path, `build-remote-sdk.mjs`, the committed `remote-sdk.mjs` bundle, CLAUDE.md, and READMEs all updated. ADRs untouched.

## Model-window clamp (`67b3183`)

A mid-session swap to a smaller-window model updated `contextWindow` but left `budget` oversized, so the 90%-of-budget compaction trigger could fire only after the real window was exhausted — exactly while Accordion suppresses pi's native compaction. main's client-side re-snap was lost when Truth moved into the extension (ADR 0021). Budget now clamps **down only** (never stomps a human's budget upward) wherever the live Truth learns its window — `model_select`, rebuilds, and the late-learned-window path (fixing the corollary where a null-at-first-build window left the 70k default forever). Aged-summary conductors also trigger on `min(budget, contextWindow)` as defense in depth. Clamp lives at the extension call sites (not inside `Truth.setContextWindow`) so replicas replay the exact same single-field config events and stay rev-locked — verified by new replica lockstep tests.

## Native compaction: narrower suppression + visible notice (`1d813f1`, `51dd4ca`)

Suppression of pi's native compaction was gated on *any* connected WS client, so a read-only viewer with folding off disabled pi's own overflow safety net. It is now gated on **folding armed** only: pi keeps its native compaction unless Accordion is actually rewriting the wire. When a native compaction runs while clients are viewing, the map rebuilds via the existing structural-divergence path and every attached UI shows a small transient notice — a new minimal server→client `notice` message (**protocol v17**), rendered as a toast styled after the existing controller-switch `DemotionToast` (same corner, panel/border/typography, auto-dismiss). The committed thermocline `remote-sdk.mjs` was regenerated for the version bump (byte-reproducible from source).

## Verification

- vitest: 707/707 (44 files) · `svelte-check`: 0 errors / 0 warnings
- `extension/smoke.mjs` + `smoke-conductor.mjs` (real out-of-process thermocline spawn): pass; smoke now covers the suppression gate, the compaction notice broadcast, and imports `PROTOCOL_VERSION` instead of stale literals
- `npm pack --dry-run`: full prepack pipeline green; tarball still excludes `conductors/`
- Every commit passed an independent adversarial review before push (replica rev-lockstep, retry-storm hypothesis, prompt byte-parity vs `origin/main`, reorg completeness, bundle drift all explicitly verified)

https://claude.ai/code/session_01KKiUoB2VbAaDQp8qeWEh7h


a-Fig:
Review verdict: **request changes**. I found two issues that should be addressed before merge.

1. **[P1] Fragmented aged groups can still make compaction grow or overflow the wire.**

   `emitCoverageGroup()` inserts the complete `this.text` summary into every contiguous run separated by a held or foreign-grouped block, while the visible-token calculation charges that summary only once. Existing human holds survive exclusive-conductor attachment, so this is reachable with ordinary pre-existing pins. A focused reproduction using three pins produced four identical groups and increased live tokens from **1,200 to 4,360**. This preserves the central failure mode the PR is intended to remove. Please use a single summary carrier, or account for every emitted copy and enforce non-growth.

   Relevant code: `conductors/in-process/agedSummaryConductor.ts`, especially the single-summary accounting around line 299 and repeated digest emission around line 415.

2. **[P2] Manual `/compact` leaves attached maps stale.**

   The new `session_compact` handler only sends the CLI/UI notifications; it does not reconcile the authoritative Truth. Pi 0.80.2's manual-compaction path saves the compaction, emits `session_compact`, and returns without a subsequent `context`, `message_end`, or `agent_end`. The app can therefore continue showing pre-compaction history indefinitely while the toast says the map was rebuilt. Please read the updated session branch and invoke the existing reconciliation path inside `session_compact` before broadcasting the notice.

   Relevant code: `extension/accordion.ts`, lines 2457–2465.

Everything else I independently ran was green: **707/707 tests**, **svelte-check 0/0**, both extension smokes, reproducible thermocline bundle, and `npm pack --dry-run`. GitHub currently reports no configured checks for this branch.

— sol5.6

a-Fig:
Both findings verified as real. Disposition per owner direction:

**P2 (manual `/compact` leaves attached maps stale) — fixed in c12dbd0.** The `session_compact` handler now reconciles the Truth immediately: it resolves the post-compaction history via `readSessionMessages` (pi's own `buildSessionContext()`) and runs it through the existing `ingestMessages` structural-divergence path — forced client resnapshot + conductor resync — *before* broadcasting the notice, so the toast's "has been rebuilt to match" is true by the time it arrives. An empty read keeps the current Truth (a best-effort failure can never blank the map; a real post-compaction branch always contains at least the compaction summary entry), the reconcile runs even with zero clients attached, and this is not the `context` hook path so the no-disk-I/O invariant is untouched. Independently re-verified green: extension bundle + both smokes, 707/707 app tests, svelte-check 0/0.

**P1 (duplicated summary in fragmented coverage groups) — tracked as #90 for a dedicated fix.** Confirmed real and reachable via pre-existing pins, with one lineage note: the per-run full-digest emission and charge-once accounting are byte-for-byte `main`'s behavior (this PR made fragmentation rarer by removing the user-block run-splitting, but inherited the growth mode itself). Since the fix deliberately goes beyond main parity — single summary carrier + accounting that matches what is actually emitted, with a non-growth invariant test — the owner opted to land it as its own change rather than fold it into this parity PR. Full mechanism, fix shapes, and acceptance criteria are in the issue.

---

PR #91, opened 2026-07-24
fix(conductors): stop duplicating the aged summary into every fragmented coverage run (#90)

### LINKED ISSUE #90: Aged-summary conductors duplicate the full summary into every fragmented group run — accounting charges it once, so compaction can grow or overflow the wire
## Summary

`AgedSummaryConductor.emitCoverageGroup()` (`conductors/in-process/agedSummaryConductor.ts`, ~line 405) emits one `group` command per maximal contiguous run of covered blocks, and **every run carries the complete summary text as its digest**. The trigger accounting (~line 299) subtracts `textTokenCost()` exactly **once**:

```ts
const savedTokens = this.text !== null ? Math.max(0, sumTokens(survivors) - this.textTokenCost()) : 0;
```

If the aged region is split into K runs, the wire carries K copies of the summary while the conductor's visible-token math believes it carries one. When the summary is larger than a run's actual content, "compaction" **grows** the wire — and the K−1 uncharged copies mean the 90%-of-budget trigger under-fires, so the conductor can also let the session overflow while believing it is under budget.

## Reachability

Human pins made **before** an exclusive conductor attaches survive attachment (the `human-steering` lock only refuses *new* steering), and each held block splits a survivor run in `emitCoverageGroup`'s walk. So this is reachable with ordinary pre-existing pins scattered through old history — no exotic setup required.

Reviewer repro (sol5.6, PR #88 review): three pins → four identical groups, live tokens **1,200 → 4,360**.

## Lineage

This is **not** a PR #88 regression — the identical per-run full-digest emission and charge-once accounting exist on `main` (`conductors/compaction-naive/compaction-naive.ts`; its own comment says every run carries the summary digest). PR #88 made fragmentation *rarer* than devmain (user blocks no longer split runs after the `includeInGroup` exclusion was removed — only pins and foreign groups do), but the growth failure mode itself is inherited from main and affects both `compaction-naive` and `handoff` through the shared base.

## Fix direction

Two viable shapes, one shared base-class change either way:

1. **Single summary carrier (preferred):** the first run gets the full summary digest; subsequent runs get a tiny continuation marker (a few tokens, e.g. `[…continued from the summary above]`). Accounting then charges the full text once plus each marker's cost — matching what is actually emitted by construction.
2. **Charge every copy + enforce non-growth:** keep the duplicated digests but subtract K × `textTokenCost()` in the accounting, and skip emitting any run whose digest costs more than the content it replaces (that content simply stays live — lossless).

Either way the invariant to enforce and test is: **applying the conductor's coverage groups never increases live wire tokens, and the trigger's `visible` computation equals the true post-fold wire size.**

## Acceptance criteria

- A pins-splitting fragmentation test (≥2 held blocks inside the aged region) asserting (a) non-growth of live tokens and (b) `visible` accounting consistent with the emitted groups.
- Both `compaction-naive` and `handoff` covered via the shared base.
- Behavior with zero fragmentation (the common single-run case) byte-identical to today.

Found in review of PR #88 (see the sol5.6 review comment there for the full repro context).

a-Fig:
## The bug

`AgedSummaryConductor.emitCoverageGroup()` — the shared base for `compaction-naive` and `handoff` — emits one `group` command per maximal contiguous run of covered survivor blocks, and every run carried the FULL summary text as its digest. Meanwhile `conduct()`'s trigger accounting charges `textTokenCost()` exactly once:

```ts
const savedTokens = this.text !== null ? Math.max(0, sumTokens(survivors) - this.textTokenCost()) : 0;
```

When a held/pinned block or a foreign group fragments the aged region into K > 1 runs, the wire carried **K copies** of the summary while the accounting believed 1 — so "compaction" could grow the wire, and the K−1 uncharged copies made the 90%-of-budget trigger under-fire (the conductor could let the session overflow while believing it was under budget).

## The fix

Only the **first** emitted run carries `digest: this.text`. Every subsequent run carries `digest: ""` — which the existing group-op contract (`core/ops.ts` / `Truth.isDropGroup`) already defines as a true wire **DROP** (no message inserted at all), not the default recap `undefined` would produce. No new representation was needed; `""` is exactly "no extra text on the wire."

## Accounting rationale

The charge-once formula was always "one full-text charge, the rest free" — the bug was that the *emission* didn't match that assumption. With later runs now genuinely free on the wire, the existing accounting became correct with **zero changes** to the trigger math. K=1 (the common, unfragmented case) is byte-identical to before: same single group, same verbatim digest, same `savedTokens`.

## Known bounded residual (pinned by test)

Two small, pre-existing accounting gaps remain, both documented in code and neither introduced here:

- The wire charges a REPLACE run `estTokens(digest) + BLOCK_OVERHEAD` (4 tokens of framing) — the conductor's `textTokenCost()` never modeled the framing, including in the K=1 case (kept as-is for byte-identity).
- The wire's role-validity floor (`computeDegradedDropRuns`, `core/wire.ts`) can degrade a dropped run into a small paid recap stub when removing it would weld two same-role neighbors together. In that case the trigger under-counts the true wire by ~one recap (~25 tokens incl. framing) per degraded run — bounded and small vs. the unbounded K× full-summary error this PR removes. The "degraded-recap residual" test pins the under-count to **exactly** one recap + BLOCK_OVERHEAD so it can never grow silently.

## Test coverage (`compaction-naive.test.ts` + `handoff.test.ts`)

- **K=2 fragmentation** — exactly one group carries the summary; the other carries the empty (DROP) digest; non-growth of live wire tokens vs. the raw baseline.
- **K=3 fragmentation** — only the FIRST (earliest) run carries the text; every later run is dropped.
- **Accounting-equals-wire** — the trigger's `visible` equals the true post-fold wire size (within the one BLOCK_OVERHEAD framing term), on a fixture where the drop is genuinely free.
- **Degraded-recap residual** — non-growth holds and the under-count is exactly one recap + BLOCK_OVERHEAD when the role floor degrades the drop.
- **K=1 regression** — zero-fragmentation output byte-identical to pre-fix (same digest, same savedTokens).
- **Handoff K=2** — proves the subclass inherits the base-class fix. (Notable finding, documented in the test: acquiring `human-steering` releases pre-existing human overrides via `Truth.releaseLockedDomains`, so a *locked* fixture cannot pin-fragment — the lock-free setup matches the compaction suite's convention; foreign-group fragmentation remains reachable under locks.)

Full suite: 44 files / 713 tests green; `svelte-check` 0 errors / 0 warnings.

Fixes #90


a-Fig:
I found two blocking issues at `ffa579f`; I don't think this is merge-ready yet.

1. **[P1] The degraded-DROP residual can accumulate enough to recreate the non-growth failure.** The ~25-token cost is bounded per degraded run, not per compaction. In a focused fixture with 101 alternating assistant/user messages (10 tokens each), every user held, the conductor emitted 51 groups (one summary carrier + 50 empty-digest groups). The role floor degraded the interior DROP runs and the wire grew from **1,010 raw tokens to 1,741 live tokens**. The trigger still undercounted the actual wire by the accumulated recap costs, so issue #90's non-growth/accounting invariant is not restored. The current one-degraded-run test uses large 100-token runs and cannot catch this.

2. **[P1] Later DROP groups can apply when the designated summary-carrier group is rejected.** `carriedText` is advanced while commands are constructed, but the group ops apply independently. I reproduced a first run containing a tool call separated from its held result: Truth rejected that first summary-bearing group as `nothing collapses (all stragglers)`, then accepted the later `digest: ""` group. The final state contained only a DROP group and no summary carrier, so the later covered text disappeared from the wire without the generated summary being present.

The fix should account for the true emitted wire cost across all degraded runs and ensure DROP groups cannot commit unless a valid summary carrier commits. Please add regressions for the many-fragment growth case and the rejected-carrier/accepted-DROP case.

The normal branch checks are green: **713/713 tests**, **svelte-check 0/0**, both extension smokes, and `npm pack --dry-run`. GitHub reports CLEAN/MERGEABLE, but those checks do not cover the two failures above.

One reachability note: the shipped conductors' live lock path clears existing human pins/groups, while the new fragmentation tests attach the conductors without applying those locks. That makes the tests useful for this shared base-class seam, but not an end-to-end proof of the claimed live-host scenario.

— sol5.6

a-Fig:
Both P1s confirmed by independent reproduction and addressed in `5cf238d` + `3f9d7ed` (+ a doc nit in `c692f62`). Thank you — both were real, and chasing them surfaced a third, deeper instance of the second one.

**P1 #1 (accumulating degraded-DROP residual).** Fixed in two layers:
- *Honest accounting:* `conduct()` no longer derives `visible = rawTotal − savedTokens`; it reads `view.liveTokens` — Truth's own number, which already prices degraded runs via `computeDegradedDropRuns`. The trigger cannot drift from the wire by construction, for this or any future degradation mechanism.
- *Growth avoidance:* a new per-run economics check (`dropEconomics`) skips proposing any DROP run whose realizable saving (removable-messageKey members only, stragglers excluded) doesn't beat its worst-case cost (one `roleFloorRecap` stub per collapsed sub-run). On your 101-fragment fixture: was 1,010 raw → 1,744 live; now 1,010 → 1,020, with the trigger seeing exactly 1,020. A boundary scan puts the skip/propose flip within one token of the real wire cost.

**P1 #2 (DROP commits while the carrier is rejected).** Your `carriedText`-advances-while-ops-apply-independently diagnosis was right, and the first fix attempt (pre-vetting each run with the carrier-viability predicate, now extracted to `core/groupShape.ts` so Truth and conductors share one implementation) turned out to be *insufficient*: `Truth.opGroup` validates `snappedRange(run)`, not the proposed run, so a human-held **sibling part of the same message** could still get the carrier clamped (`human-override`) while the sibling DROP committed — same data loss, one level deeper. Final fix follows thermocline's `safeRunFromUnits` precedent: `snapToMessageAtoms` trims every candidate run inward to a fixed point of the snap, then vets and proposes exactly that set. Proposed == applied, byte-identical; a 900-pass randomized fuzz over pinned sibling parts / split tool pairs / protected boundaries produced zero clamps and zero drops-without-carrier.

**Regressions added:** your many-fragment growth fixture, the rejected-carrier shape (held sibling part), the tool-call-sibling snap-widening shape, straggler/sub-run pricing, a boundary-equality case, and a paid-retry back-off test — honest accounting exposed that a conductor stuck over trigger with nothing droppable would relaunch a billed `host.complete()` every turn; a measured back-off gate (re-opens on genuine refill) now bounds it (220 turns: 7 calls, was 220).

**On your reachability note:** agreed, and worth stating plainly — the shipped conductors' lock path (`releaseLockedDomains` on attach) clears pre-existing pins, and `coveredIds` is prefix-only, so K > 1 is reachable today only through the shared base-class seam, not on a live locked session. This PR's hardening is defensive for the base class and for any future subclass that doesn't lock `human-steering` — the tests exercise the seam at the same level as the fixtures in this thread.

726 → 733 tests, all green; svelte-check clean. The four new regression tests were verified to fail against the pre-fix revisions.

a-Fig:
Delaying. Not essential.

a-Fig:
Final merge review against current `devmain`: one integration blocker was reproduced and fixed in `bf3ff64`. With provider calibration above 1, the DROP economics mixed calibrated block tokens with an uncalibrated framing cost and could approve a raw-wire-growing DROP (probe at `k=2`: 24 tokens removed, 25-token recap inserted). The merge resolution now uses the original wire estimate only for this structural sign check, while all trigger/budget decisions remain calibrated; the exact repro is pinned by test.

The resolution also preserves Triptych’s top-band boundary and #99’s sticky skeleton-init failure behavior, removes obsolete manual skeleton-savings reconstruction in favor of authoritative `view.liveTokens`, and regenerates both remote SDK bundles.

Final verification: 801/801 tests, `svelte-check` 0 errors / 0 warnings, extension smoke + real conductor spawn smoke, production app/package build, and `npm pack --dry-run` all green. No remaining review blockers.

PR #98, opened 2026-07-25
feat(#93): show the system prompt — never foldable, counted in totals

### LINKED ISSUE #93: Show everything actually sent to the model, including the system prompt
## What

Accordion's Truth/Map/Transcript views only render the `Block` kinds captured off pi's message hooks (`user · text · thinking · tool_call · tool_result` — `core/types.ts`). Anything else that rides the wire to the model on a given call is invisible: most notably the **system prompt**, but potentially other request-level content pi attaches outside the message array (e.g. a custom system prompt injected via `before_agent_start`, or other out-of-band context).

The ask: everything that is actually included in the model's context for a call should be visible and inspectable in Accordion — not just the message-log blocks it tracks today.

## Why

Accordion's whole pitch is "see and control what the agent sees." A user auditing budget/token usage or debugging a weird response currently has no way to see the system prompt's token cost or content inside the tool, even though it's part of every single call and can be substantial. That's a blind spot in the exact thing this product exists to visualize.

## Notes for implementation

- The extension already has a hook that sees the full outgoing payload (`context`, and `before_provider_request` sees the fully serialized provider payload) — one of these is presumably where the system prompt would need to be captured and forwarded into the `Truth`/wire protocol so a replica can render it.
- Needs a design decision: does the system prompt become a new `Block` kind (with its own color in the visual grammar table), or a separate always-pinned pseudo-block/header treated specially (never foldable/foldable-with-restrictions, never counted against the protected tail the same way)? It changes per-session rarely, so it likely shouldn't be dice-graded turn-by-turn like a normal block.
- Whatever shape it takes must still respect the **RULE** in CLAUDE.md — read-only/CC/preview modes can't show anything the live steering path couldn't also show, so this needs to flow through the same `Truth`/protocol path as everything else, not be bolted on as a client-only special case.
- Token accounting (`core/tokens.ts`, budget math) presumably needs to include it too, since right now the displayed budget usage silently excludes system-prompt tokens.

## Scope

Exact mechanism (new Block kind vs. dedicated header treatment) is an open design question — flagging the gap here rather than prescribing the implementation.

a-Fig:
Replacement review PR for #96 after #97 reverted its premature merge.

This reapplies the original #96 commit unchanged on top of the latest `devmain` so the feature can receive a deliberate review before merging again.

Original author: @shahxsheel
Closes #93.

### DIFF
--- a/core/conductor/hostAdapter.ts
+++ b/core/conductor/hostAdapter.ts
@@ -104,12 +104,13 @@ export function hostEventsFromTruthEvent(truth: Truth, e: TruthEvent): HostEvent
 		return changes.length ? [{ type: "state-changed", changes, rev: e.rev }] : [];
 	}
 	if (e.type === "config") {
-		// `calibration` (v18, issue #11 stage 1) and `systemPrompt` (v19, issue #93) are both
-		// DISPLAY-only (a conductor reads `systemPrompt` via `ConductorHost.systemPrompt()` directly
-		// instead) and must stay invisible as a `state-changed` notification — neither carries a
-		// `budget`/`protectTokens`/`contextWindow` field, so without this guard either would fall
-		// through to the `budget !== undefined ? "budget" : "protect"` default and get mislabeled a
-		// "protect" change, waking every subscribed conductor for a dial it was never meant to see.
+		// A system-prompt change moves `stats().liveTokens`/`fullTokens`, so it is decision-bearing
+		// for ViewConductors even though the prompt itself is read-only. Surface a distinct change
+		// rather than leaving compaction/handoff planned against the previous total until the next
+		// turn. Calibration remains host/display plumbing and intentionally emits no HostEvent.
+		if (e.systemPrompt !== undefined) {
+			return [{ type: "state-changed", changes: [{ what: "systemPrompt", by: "you" }], rev: e.rev }];
+		}
 		if (e.budget === undefined && e.protectTokens === undefined && e.contextWindow === undefined) return [];
 		const what: StateChange["what"] = e.budget !== undefined ? "budget" : "protect";
 		return [{ type: "state-changed", changes: [{ what, by: "you" }], rev: e.rev }];
--- a/core/conductor/hostAdapter.ts
+++ b/core/conductor/hostAdapter.ts
@@ -104,11 +104,12 @@ export function hostEventsFromTruthEvent(truth: Truth, e: TruthEvent): HostEvent
 		return changes.length ? [{ type: "state-changed", changes, rev: e.rev }] : [];
 	}
 	if (e.type === "config") {
-		// `calibration` (v18, issue #11 stage 1) is DISPLAY-only and must stay invisible to a
-		// conductor — it carries no `budget`/`protectTokens`/`contextWindow` field, so without this
-		// guard it would fall through to the `budget !== undefined ? "budget" : "protect"` default and
-		// get mislabeled a "protect" change, waking every subscribed conductor on every calibration
-		// snap (once per model reply) for a dial it was never meant to see.
+		// `calibration` (v18, issue #11 stage 1) and `systemPrompt` (v19, issue #93) are both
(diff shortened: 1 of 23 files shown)


core/protocol.ts
```diff
@@ -97,13 +97,26 @@
...
+ *    pre-v19 peer (which has none of this vocabulary) cannot pair with a v19 host/client that
+ *    assumes it.
  */
 import type { Actor, Group, Override } from "./types";
 import type { LockName } from "./locks";
 import { sanitizeOps, type Op, type OpResult } from "./ops";
 
 /** Bump on any breaking change to the message shapes below. */
-export const PROTOCOL_VERSION = 18;
+export const PROTOCOL_VERSION = 19;
```
a-Fig, line 119: [P1] Regenerate every committed remote SDK for protocol v19. This bump updates the host/source and the thermocline bundle, but `conductors/ws/triptych/triptych-sdk.mjs` remains at `PROTOCOL_VERSION = 18` on this branch. `npm pack --dry-run` rewrites that tracked bundle by 60 additions/7 deletions; without committing the regenerated triptych SDK, the shipped triptych runner rejects the v19 host during its hello gate.

extension/accordion.ts
```diff
@@ -505,6 +506,12 @@ export default function accordionLive(pi: ExtensionAPI, dependencies: RuntimeDep
 	let model = "";
 	let tokens: number | null = null;
 	let contextWindow: number | null = null;
+	// Issue #93: module-level cache of the most recently captured system prompt, same shape as
+	// `contextWindow` above and for the same reason — `refreshFromCtx` runs before `truth` may be
+	// null/about to be rebuilt (`ingestMessages` → `buildTruth`), so the captured value must survive
+	// that gap and get re-applied once a fresh `Truth` exists (see `buildTruth`'s re-apply).
+	let systemPromptText: string | null = null;
```
a-Fig, line 513: [P1] Clear this cache at every `session_start`. `truth` and the calibration pairing are reset there, but `systemPromptText`/`systemPromptTokens` are not. Because `refreshFromCtx` is explicitly best-effort and one exception from an optional context API aborts the rest of that function, `buildTruth` can apply the previous session's prompt to a new session. That leaks prompt content across sessions and corrupts the new session's token totals. Reset both fields before the session-start refresh and cover the failure/absence path with a two-session test.

core/conductor/hostAdapter.ts
```diff
@@ -104,11 +104,12 @@ export function hostEventsFromTruthEvent(truth: Truth, e: TruthEvent): HostEvent
 		return changes.length ? [{ type: "state-changed", changes, rev: e.rev }] : [];
 	}
 	if (e.type === "config") {
-		// `calibration` (v18, issue #11 stage 1) is DISPLAY-only and must stay invisible to a
-		// conductor — it carries no `budget`/`protectTokens`/`contextWindow` field, so without this
-		// guard it would fall through to the `budget !== undefined ? "budget" : "protect"` default and
-		// get mislabeled a "protect" change, waking every subscribed conductor on every calibration
-		// snap (once per model reply) for a dial it was never meant to see.
+		// `calibration` (v18, issue #11 stage 1) and `systemPrompt` (v19, issue #93) are both
+		// DISPLAY-only (a conductor reads `systemPrompt` via `ConductorHost.systemPrompt()` directly
```
a-Fig: [P1] Prompt changes are not display-only anymore. `Truth.liveTokens()`/`stats()` now include this value, and the PR explicitly supports mid-session prompt drift, but this guard drops the only Truth event that could wake an attached ViewConductor. A prompt expansion captured immediately before a model call can therefore leave compaction/handoff planned against the old total until after that call, potentially missing the budget boundary. Surface a distinct conductor-visible state change (or equivalent resync) for system-prompt changes instead of suppressing it.

a-Fig (commented):
Review found three blocking correctness issues. The feature-level tests and package smoke suite pass, but the committed package state is not integration-safe on current `devmain`, and two lifecycle paths can expose stale prompt state or leave conductor decisions stale. Treat the inline P1 findings below as changes required before merge.

PR #103, opened 2026-07-31
fix(#102): stabilize calibrated context totals

### LINKED ISSUE #102: Context value # sometimes spikes randomly on devmain.
I was watching accordion run and as the agent was working the token count int accordion randomly went from 24k to 71k back down to 27k or so. Very strange.

a-Fig:
Closes #102.

## What changed

- replace the single global token multiplier with an affine calibration model: `real = base + scale × estimate`
- fit the scale over the eight most recent provider observations, then re-anchor the base through the latest receipt so the newest real total remains exact
- apply fixed overhead once to whole-context totals while using only the scale for per-block/component deltas
- carry the affine base through Truth snapshots/config events and bump the live protocol to v20
- update the UI's headline, budget flag, composition axis, and anchored marker to use the affine total
- regenerate both committed remote-conductor SDK bundles and document the issue #102 addendum in ADR 0025

## Root cause

The prior calibration used `k = real / estimated` and multiplied every current block by `k`. Fixed system/tool-schema overhead was therefore smeared across blocks. In the reported session, the last receipt before a large tool result produced an overhead-heavy ratio. Appending the ~8.8k-token tool result caused that old ratio to be applied to content the receipt had never observed, projecting about 71.5k tokens. The next provider receipt recomputed `k` and immediately corrected the same context to about 24.9k.

The affine model separates fixed request overhead from content growth. A new tool result now changes the estimate by the fitted content slope instead of inheriting the fixed overhead repeatedly.

## Validation

- `npm run check` — 0 errors, 0 warnings
- `npm run test -- --run` — 785 passed, 20 skipped
- `node smoke.mjs` — passed
- `node smoke-conductor.mjs` — passed
- `npm pack --dry-run` — production app build, client copy, extension bundle, both smoke suites, remote SDK generation, and tarball listing all passed
- `git diff --check` — passed


a-Fig:
Closing this implementation so issue #102 can be redone from scratch with calibration scoped only to blocks covered by each provider receipt.

PR #104, opened 2026-08-03
fix(#102): scope token calibration to measured blocks

### LINKED ISSUE #102: Context value # sometimes spikes randomly on devmain.

a-Fig:
## What changed

- Capture the last block order together with each departing-wire estimate.
- Apply the resulting provider multiplier only to blocks covered by that receipt; later assistant and tool blocks remain raw until the next receipt advances the frontier.
- Use the same covered/raw basis in headline totals, block/group readouts, protected-tail sizing, and conductor views.
- Reset calibration on structural rebuilds where an order frontier can no longer prove coverage.
- Carry the frontier through protocol v20 snapshots/events and regenerate the remote conductor SDKs.

## Root cause

The provider multiplier described one completed request, but Accordion treated it as a timeless multiplier for the entire current block log. In the captured issue trace, 2,539 estimated tokens paired with 15,943 real input tokens (about 6.28x). An 8,851-token tool result then arrived before the next receipt, and the stale multiplier projected the total to about 71.5k. The following receipt replaced k and the display dropped back near 25k. The stale installed extension amplified the ratio by omitting newer system-prompt accounting, but current devmain still had the same category error.

The corrected aggregate is coveredRaw * k + newerRaw, not (coveredRaw + newerRaw) * k.

## Verification

- npm run check: 0 errors, 0 warnings
- npm run test -- --run: 802 passed
- node smoke.mjs: passed
- node smoke-conductor.mjs: passed
- npm pack --dry-run: production build, both smokes, and tarball inspection passed
- Regression test reproduces the old ~71.5k projection and asserts the new total remains 24,794 until the next receipt covers the new block

Closes #102


app/src/lib/ui/map/ContextMap.svelte
```diff
@@ -388,16 +388,16 @@
 	});
 
 	const k = (n: number) => { n = Math.round(n); return n >= 1000 ? `${(n / 1000).toFixed(n >= 10000 ? 0 : 1)}k` : `${n}`; };
-	// issue #11 stage 1: "≈" marker gate for the transcript's visible token badge — `calibration === 1`
+	// issue #102: "≈" marker gate for the transcript's visible token badge — a null receipt frontier
 	// covers BOTH cold start (no observation yet) and every read-only/demo/CC/file session (no live
 	// host ever calibrates those), so no separate `readOnly` prop plumbing is needed here.
-	const notAnchored = $derived(store.calibration === 1);
+	const notAnchored = $derived(store.calibrationThroughOrder === null);
```
a-Fig, line 394: **The "≈" marker doesn't follow the frontier — recommend fixing before merge.**

Moving the gate from `calibration === 1` to `calibrationThroughOrder === null` is a strict improvement, but it's still *one boolean for the whole session*. The premise of this PR is that per-block coverage is now heterogeneous — so after the first receipt lands, `notAnchored` goes false and every block **past** the frontier renders its raw chars/4 estimate **bare, with no "≈"**. Those are precisely the blocks this PR just declared unanchored.

Same issue at `Inspector.svelte:82` and `MapHeader.svelte:76`.

`Truth` already has exactly the predicate needed — `isCalibrated(b)` — it just isn't re-exported through `AccordionStore` alongside the new `calBlockTokens`/`calGroupFullTokens` helpers. Per-block would be:

```svelte
{#if !store.isCalibrated(b)}<span class="approx" aria-hidden="true">≈</span>{/if}
```

For the `MapHeader` hero the total is now a *mix* of anchored and unanchored, so arguably it should show "≈" whenever anything is uncovered (`calibrationThroughOrder === null || calibrationThroughOrder < lastOrder`).

This is small, and it's the user-visible half of the fix — without it the UI asserts more confidence than the new model actually has.

---

CLAUDE.md
```diff
@@ -56,7 +56,7 @@ Guidance for AI coding sessions. [VISION.md](VISION.md) = product north star ·
...
+- `Truth.calibration`/`calibrationThroughOrder` — a host-set, rev-stamped provider anchor (`k = realTokens/estTokens`) plus the last block order on the exact wire that receipt measured. Covered blocks use `calBlockTokens`; blocks appended afterward stay raw until a later receipt advances the frontier (issue #102). `protectedFromIndex()` and `stats()` use the same mixed covered/raw basis, and every conductor decision read (`ViewBlock.tokens`/`foldedTokens`, `ConductorHost.countTokens`) stays unit-consistent — `budget`/`protectTokens`/`contextWindow` remain literal real-token dials. `ViewBlock.rawTokens` is the narrow structural-sign exception. See [ADR 0025](docs/adr/0025-token-calibration.md)
```
a-Fig, line 59: **`ConductorHost.countTokens` is not actually unit-consistent any more — this line overstates it.**

`countTokens` still routes through the *global* `calTokens` in all three implementations:

- `core/conductor/liveHost.ts:222` — `return t ? t.calTokens(raw) : raw;`
- `core/conductor/remote.ts:239` — `return replica!.calTokens(estTokens(text));`
- `core/conductor/testhost.ts:68` — `return this.truth.calTokens(estTokens(text));`

Meanwhile `ViewBlock.tokens` is now `calBlockTokens`, i.e. **raw** for any block past the frontier. So a conductor doing the standard "will my replacement summary be smaller than what it replaces?" check compares a k-scaled number against a raw one. At the issue's own `k ≈ 6.28` that's a ~6× error, landing on exactly the newest blocks conductors act on.

Leaving it as-is is defensible — a bare string has no block identity, so there's no frontier to consult — but it should be a stated decision rather than an inherited one, and this line shouldn't claim consistency that the code no longer provides. Options: scale it, expose a block-relative variant for the substitution check, or reword here + add a note to ADR 0025's Consequences.

---

core/truth.ts
```diff
@@ -330,7 +350,13 @@ export class Truth {
...
 		next.holderLabel = prev.holderLabel;
 		next.activeTailTok = prev.activeTailTok;
-		next.calibrationMul = prev.calibrationMul;
+		// A structural rebuild can insert/reorder blocks under stable order values. A scalar coverage
+		// frontier can no longer prove which representations the last receipt actually covered, so
+		// return to a cold estimate until the request departing from this rebuilt Truth gets a receipt.
+		// Carrying k here would recreate issue #102 for a newly inserted block below the old frontier.
+		next.calibrationMul = 1;
+		next.calibrationThroughOrderValue = null;
+		next.systemPromptCalibratedValue = false;
```
a-Fig, line 359: **Hard-resetting to cold is defensible, but the better pattern lives ~30 lines below.**

The reasoning is sound — an order-only scalar can't prove coverage after inserts/reorders. But `sentThroughOrder` solves the *identical* problem in `rebuildFrom` (lines ~384–401) without going cold: it drags the order-prefix frontier back to just before the earliest not-previously-present block, and keeps a per-id `carriedSent` set for the survivors that fall below it. The same treatment would preserve calibration coverage for every block that demonstrably survived the rebuild.

As written, the consequence is a visible readout jump in the *other* direction: after any structural divergence the display drops from `covered·k` back to raw chars/4 and re-shows "≈" until the next receipt. With `k ≈ 6×` early in a session that's a ~6× downward step — the same family of surprise as #102, just rarer and self-healing in one model call.

Divergence isn't as rare as "compaction / fork / tree-nav" suggests, either — `accordion.ts:1891` documents that a second GUI attaching mid-session can spuriously trip the check.

I'd accept this for v1, but: worth a line in ADR 0025's Consequences, and reusing the `carriedSent` pattern is the obvious follow-up.

---

core/truth.ts
```diff
@@ -618,28 +685,18 @@ export class Truth {
...
-		const cap = target * PROTECT_OVERFLOW_CAP;
-		let sum = blocks[blocks.length - 1].tokens;
-		if (sum >= target) return blocks.length - 1;
+		// Covered blocks use k; blocks newer than the receipt frontier stay raw. Use unrounded float
+		// costs during the walk so cumulative per-block rounding cannot move the boundary. Host and
+		// replica carry the same k/frontier and block order, so this is deterministic on both sides.
+		const realCost = (b: Block) => b.tokens * (this.isCalibrated(b) ? this.calibrationMul : 1);
+		const cap = targetReal * PROTECT_OVERFLOW_CAP;
+		let sum = realCost(blocks[blocks.length - 1]);
+		if (sum >= targetReal) return blocks.length - 1;
```
a-Fig, line 694: **Correct, and the determinism argument survives — but the tail boundary now moves every time the frontier advances, which isn't documented.**

The determinism concern the old comment argued at length is genuinely preserved: the old worry was `Math.round` *inside* the walk, and this version keeps unrounded float costs over an identically-ordered array on both sides, so there's no accumulation to diverge on. Good.

What's new is that the boundary is now a function of frontier position, not just of `k`. I measured it (40 × 1,000-raw-token blocks, `protectTokens` 20k):

```
k=1                  → pfi 20   (20 tail blocks)
k=6.28, frontier=29  → pfi 28   (12 tail blocks)
k=2.17, frontier=39  → pfi 30   (10 tail blocks)
```

So during a high-k early session the fold gate's boundary wobbles per turn as blocks flip from weight 1 to weight k. `setCalibration` runs `housekeep()`, so nothing goes stale — and net this is *more* protective than the old `targetReal / k` division, which at k=6.28 would have collapsed the tail to ~3 blocks. So I read it as an improvement, not a bug.

But it is an unadvertised behavior change. ADR 0025's rewritten `protectedFromIndex` section explains the mechanism without noting that the boundary now churns per receipt — worth a sentence there.

---

core/replica.ts
```diff
@@ -152,6 +160,8 @@ export function wireEventFromTruthEvent(e: TruthEvent): WireEvent | null {
 				contextWindow: e.contextWindow,
 				protectTokens: e.protectTokens,
 				calibration: e.calibration,
+				calibrationThroughOrder: e.calibrationThroughOrder,
+				systemPromptCalibrated: e.systemPromptCalibrated,
```
a-Fig, line 164: **`systemPromptCalibrated` is emitted on `config` events but never consumed — dead field on the event path.**

`applyWireEvent` (line ~191) reads only `calibration` + `calibrationThroughOrder`; the replica then re-derives the flag inside `setCalibration` as `this.systemPromptText !== null`.

That's correct *today*, but only because event ordering guarantees the host's `config{systemPrompt}` is replayed before its `config{calibration}` — so both sides evaluate the same predicate. It's a real coupling that nothing asserts, and it silently becomes a host/replica fork if the emission order ever changes.

It genuinely *is* needed on `SnapshotState` (no ordering to lean on there), so no issue with that half. Suggest either consuming it in `applyWireEvent` (pass it through to `setCalibration`) or dropping it from `TruthEvent`/`WireEvent`'s config shape. An emitted-but-ignored wire field is the kind of thing a future change trusts by accident.

---

app/src/lib/ui/map/MapHeader.svelte
```diff
@@ -458,13 +458,13 @@
 		<div class="bar" bind:this={barEl} role="img" aria-label="Context composition">
 			{#each LADDER as seg (seg.kind)}
 				{@const v = liveByKind[seg.kind]}
-				{@const calV = store.calTokens(v)}
+				{@const calV = v}
```
a-Fig, line 461: Nit: `{@const calV = v}` is now a vestigial no-op alias — `liveByKind` already calibrates per block at line 53, so `calV === v` by construction. Inline `v` at both use sites and drop the `{@const}`.

Same shape two lines down: `{@const calSaved = store.calSavedTokens}` inside `{#if store.calSavedTokens > 0}` is just a rename.

---

app/src/lib/ui/map/ContextMap.svelte
```diff
@@ -1093,7 +1093,7 @@
 												{#if item.kind === "tile"}
 													{@const b = item.block}
 													<div
-														class="cell face f{faceFor(b.tokens)} k-{b.kind}"
+												class="cell face f{faceFor(b.tokens, b)} k-{b.kind}"
```
a-Fig, line 1096: Nit: indentation regression — this line lost two tabs relative to its sibling attributes (`class:sel`, `class:pinned`, … are at one level deeper than the `class=` it belongs to).

Same at lines **1112**, **1128**, and **1164**. Purely cosmetic, but all four are `class="cell face f{…}"` lines touched by this diff, so it's a consistent slip worth a quick re-indent.

---

app/src/lib/ui/map/ContextMap.svelte
```diff
@@ -74,7 +74,7 @@
 	// what the rest of the UI now reports as that block's real size, not the raw chars/4 estimate.
 	// Wrapped once here (rather than at each call site) so `faceFor(x)` stays a drop-in raw-tokens-in
 	// call everywhere it's already used; the canvas/sprite machinery `faceForLib` feeds is untouched.
-	const faceFor = (tokens: number) => faceForLib(store.calTokens(tokens));
+	const faceFor = (tokens: number, block?: Block) => faceForLib(block ? store.calBlockTokens(block, tokens) : tokens);
```
a-Fig, line 77: Nit: the comment directly above (lines 74–76) is now stale — it still says `faceFor(x)` "stays a drop-in raw-tokens-in call everywhere it's already used."

That's no longer true. The optional second arg splits it into two modes:
- `faceFor(raw, block)` → raw tokens, calibrated against `block`'s coverage
- `faceFor(x)` → **already real**, passed straight through

Both are used: lines 97/152/197/1063/1096/1174 take the block form, while 1112/1128/1164 pass pre-calibrated values. All call sites are correct today (I checked each), but a bare `faceFor(b.tokens)` added later would now silently skip calibration instead of applying it — the exact opposite of the old contract. Worth updating the comment to name the two modes.

---

a-Fig (commented):
## Review: right fix, well implemented, one thing I'd want before merge

I read #102, the full 18-file diff, and re-ran the verification independently rather than taking the PR body's word for it.

### Was this the right fix? Yes.

The diagnosis is correct and the fix is at the right layer. The old model treated `k = real/est` as a timeless property of the session, but `k` describes exactly **one completed request**. Anything appended between a receipt and the next one got scaled by a multiplier that had never measured it — at `k ≈ 6.28`, an 8,851-token tool result projected as ~55k. That's the spike; the next receipt's correction is the drop back to 27k.

`coveredRaw·k + newerRaw` is right, and it's stronger than it first looks. If the real relationship is additive (`real = est + C`, C being tool-schema overhead chars/4 can't see), then `k = (E+C)/E`, and at each receipt the display evaluates to exactly `E + C` — correct — while incrementing raw in between. **The frontier fix makes the aggregate self-anchoring without needing the deferred affine model.** Declining the affine regression here is the right call and the ADR addendum says so cleanly.

The two calls I'd have worried about are both made correctly:
- **Not reusing `sentThroughOrder`** — sent-ness advances in `finally` when the wire departs; the calibration frontier has to stay pinned to the exact wire measured. Captured separately at the `context` hook, with a comment explaining why.
- **Atomic `(k, frontier)` on hydration** — `replica.ts` falls back to `k=1` when the frontier is absent rather than reviving a global multiplier from a stale snapshot.

### Verification (run, not trusted)

| check | result |
|---|---|
| `npm run check` | 401 files, **0 errors / 0 warnings** |
| `npm run test -- --run` | **802 passed** |
| `node smoke.mjs` | OK |
| `node smoke-conductor.mjs` | OK |
| regenerate both remote SDKs | **byte-identical** to committed — no generated-artifact drift |

I also grepped every UI token read surface for stragglers — the conversion to `calBlockTokens`/`calGroup*Tokens` is complete and consistent. The regression test reproduces the exact issue trace and asserts 24,794 instead of ~71.5k.

### Findings (8 inline)

**Recommend before merge:**
1. **The "≈" marker doesn't follow the frontier.** Still one session-wide boolean, so after the first receipt every block *past* the frontier shows a raw chars/4 estimate bare. `Truth.isCalibrated(b)` already exists; it just isn't exposed on the store. This is the user-visible half of the fix.

**Needs a decision, doesn't have to block:**

2. **`ConductorHost.countTokens` still uses the global `k`** (`liveHost.ts:222`, `remote.ts:239`, `testhost.ts:68`) while `ViewBlock.tokens` is now raw past the frontier — a ~6× apples-to-oranges in the "is my summary smaller?" check at k≈6.28. The new CLAUDE.md line claiming it "stays unit-consistent" overstates it.

**Follow-ups:**

3. `rebuildFrom` hard-resets to cold, where `sentThroughOrder`'s `carriedSent` pattern (30 lines below) would preserve coverage for surviving blocks. Causes a ~6× *downward* jump after any divergence — rarer, self-healing in one model call, but same family of surprise as #102.
4. The protected-tail boundary now moves each time the frontier advances (measured: 20 → 12 → 10 tail blocks). `housekeep()` runs so nothing goes stale, and it's *more* protective than the old `targetReal / k` division — just undocumented.
5. `systemPromptCalibrated` is emitted on `config` events but never consumed; the replica re-derives it, correct only because of event ordering nothing asserts.

**Nits:** 6) vestigial `{@const calV = v}` · 7) indentation regression on 4 lines in `ContextMap.svelte` · 8) stale `faceFor()` contract comment.

### Verdict

**Merge-ready once #1 is in.** #2 deserves an explicit decision. Everything else is follow-up.

Solid work — the reasoning about *why* `sentThroughOrder` can't substitute for the frontier, and the atomic `(k, frontier)` hydration fallback, are both easy to get wrong and were gotten right.

---

PR #120, opened 2026-08-07
Add button to open selected block as .md file in external program

### LINKED ISSUE #119: allow any block to be opened as a .md file in a separate program
Just add a simple button after selecting a block to view it in full as a .md file. 
It should be a tmp file to not take up permanent storage on the users device. 

It should open in VS code or notes or wherever the user usually reads .md files.

RawNuke:
Write the selected block's text to a temp .md file and open it with the OS default application via tauri-plugin-opener and tauri-plugin-fs. The button appears in the Inspector panel alongside Fold and Pin and only renders in the desktop (Tauri) environment. Closes #119.


a-Fig:
So cool, i'll review this tonight and merge it. Thank you!

a-Fig:
im sorry but this wasnt really what I had in mind so I built the feature myself.

PR #143, opened 2026-09-14
Land the editable folded digest on devmain

### LINKED ISSUE #140: Land the editable folded digest on devmain (reverted off main, re-landed on vibe as #139)
## What

Land the **editable folded digest** feature on `devmain`.

Clicking a folded block's digest text opens a real `<textarea>` so a human can author the exact string the model receives, with **Save** / **Cancel** / **Auto digest** and a live token-cost readout. Group summaries are editable the same way. Reachable from both Transcript view (the digest text on the card) and Map view (the Inspector's digest callout).

## Why this is an issue and not just a merge

The work was written and reviewed, then **merged into `main` by mistake** — on the assumption that `main` and `devmain` were equivalent, which they are not (`main` is the release trunk; `devmain` is the active development branch — CLAUDE.md, "Branching & PR workflow").

- The `main` merge (#138) has been **reverted** — `e3eb9bb`.
- The same three commits, rebased onto `vibe`, are open as **#139** (that is the branch currently being worked on).
- `devmain` never received it, so it still needs to land there.

## Scope

Cherry-pick / rebase the three commits from `claude/editable-folded-digest-vibe` onto the latest `origin/devmain` and open a PR targeting `devmain`:

- `Let humans write the digest a folded block shows the agent`
- `Fix review findings on the editable digest`
- `Move the group tag-strip from the widget into Truth`

Files touched (15, +1180 / −39):

```
CLAUDE.md
app/src/lib/engine/store.digest.test.ts   (new)
app/src/lib/engine/store.svelte.ts
app/src/lib/live/plan.groups.test.ts
app/src/lib/ui/map/ContextMap.svelte
app/src/lib/ui/map/DigestEditor.svelte    (new)
app/src/lib/ui/map/Inspector.svelte
app/src/lib/ui/map/digestDrafts.ts        (new)
conductors/ws/thermocline/remote-sdk.mjs  (regenerated)
conductors/ws/triptych/triptych-sdk.mjs   (regenerated)
core/agentView.ts
core/digest.ts
core/humanDigest.test.ts                  (new)
core/ops.ts
core/truth.ts
```

## Contract to preserve on the rebase

The load-bearing invariant is **the tag is the handle**: `{#<code> FOLDED}` is the only way a fold code reaches the model, so a *human-authored* digest is deliberately **untagged** — the agent cannot `unfold` or `recall` a block whose digest a human wrote. That is enforced inside `Truth` (the human branches of `opFold`/`opGroup` run `stripFoldTags` to a fixed point), **not** in the editor, because the editor seeds its box with the current digest and so edited engine text still starts with a tag. Strategy branches deliberately do **not** strip. Also own-tag, not any-tag — a foreign tag must not make a block reachable by its own code.

Two other details worth not losing:

- Emptying the box is asymmetric on purpose: a **block** becomes the `{empty}` sentinel (an empty `digestText` makes `applyPlan` ship the block whole — the UI lie this repo forbids), a **group** becomes a real DROP (`Group.digest === null`).
- Unsaved text is client-local (`digestDrafts.ts`) and survives a refused save, so no `SnapshotState` field and no protocol bump — still v22.

## Regeneration + verification

Because `core/` changes, both committed remote-SDK bundles must be regenerated in the same PR:

```bash
node extension/build-remote-sdk.mjs
```

Gate to re-run against `devmain` (the `main`/`vibe` runs are not transferable — `devmain` has diverged):

```bash
cd app && npm run test && npx svelte-check --tsconfig ./tsconfig.json
cd extension && node smoke.mjs && node smoke-conductor.mjs
```

## Refs

- Reverted merge on `main`: #138 → `e3eb9bb`
- The `vibe` re-land: #139

a-Fig:
## What

Lands the **editable folded digest** feature on `devmain`: clicking a folded block's digest text opens a real `<textarea>` so a human can author the exact string the model receives, with Save / Cancel / Auto digest and a live token-cost readout. Group summaries are editable the same way, from both Transcript view and the Map Inspector.

Closes #140

## Background

- This work was originally merged into `main` by mistake as #138, then **reverted** there (`e3eb9bb`) — `main` is the release trunk, `devmain` is the active dev branch, they are not equivalent.
- The same commits were separately rebased onto `vibe` and merged as **#139**.
- `devmain` never received the feature — this PR lands it there properly, cherry-picked from `claude/editable-folded-digest-vibe` onto latest `origin/devmain` (`089de50`).

## Commits

Cherry-picked in order, authorship preserved, no squashing:

1. `d4c1709` — Let humans write the digest a folded block shows the agent
2. `0ab23c3` — Fix review findings on the editable digest
3. `8c56a0f` — Move the group tag-strip from the widget into Truth
4. `593b5e8` — Fix review follow-ups on the editable digest

The 4th commit is **not named in issue #140's body** (it was pushed ~40min after the issue was filed), but it's part of the same PR #139 and fixes real bugs in the same invariants below, so it's included as in-scope.

All four applied with **zero conflicts** — no manual resolution was needed.

## SDK bundles

`conductors/ws/thermocline/remote-sdk.mjs` and `conductors/ws/triptych/triptych-sdk.mjs` were **regenerated fresh** against `devmain`'s actual `core/` state (`node extension/build-remote-sdk.mjs`) rather than reusing the versions the cherry-pick brought in from `vibe`. Regeneration produced output that is **bit-identical** to the cherry-picked bundles (verified via `git hash-object` matching the committed blob hashes) — no additional commit was needed.

## Invariants preserved (verified, not just "tests passed")

- **Tag is the handle, own-tag only**: `{#<code> FOLDED}` is the only way a fold code reaches the model. `core/truth.ts`'s `opFold`/`opGroup` human branches run `stripFoldTags` to a fixed point (`core/truth.ts:1433`, `:1639`); the editor itself does not strip. Reachability is decided by `hasOwnFoldTag` (`core/digest.ts:157-160`), not the weaker `hasFoldTag`, so a foreign/pasted tag cannot make a block reachable by its own code — covered explicitly by `core/humanDigest.test.ts`'s "a FOREIGN tag does not make a block reachable by its own code" case.
- **Asymmetric empty semantics**: an emptied block digest commits the `EMPTY_DIGEST` sentinel `"{empty}"` (`app/src/lib/engine/store.svelte.ts:431`, `core/digest.ts:198`) since `applyPlan` would ship an empty-digest block whole; an emptied group digest commits a real DROP, `Group.digest === null` (`app/src/lib/engine/store.svelte.ts:533`, consumed by `core/truth.ts`'s `isDropGroup` at `core/truth.ts:765`).
- **`digestDrafts.ts` stays client-local**: plain in-memory `Map`, no new `SnapshotState` field. `core/protocol.ts`'s `PROTOCOL_VERSION` is unchanged at `22`.

## Test gates

Fresh worktree — `npm install` run in both `app/` and `extension/` (not previously installed):

- `cd app && npm run test` — **51 test files passed, 1 skipped; 842 tests passed, 20 skipped** (0 failed). Includes `core/humanDigest.test.ts` (24 tests) and `app/src/lib/engine/store.digest.test.ts` (11 tests).
- `cd app && npx svelte-check --tsconfig ./tsconfig.json` — **0 errors, 0 warnings** (457 files).
- `cd extension && node smoke.mjs` — **OK**.
- `cd extension && node smoke-conductor.mjs` — **OK**.

No devmain-drift fixes were needed — all four commits applied and passed cleanly against devmain's current code shape. (One environment-only step was needed before tests would run at all: `npx svelte-kit sync` to generate the gitignored `.svelte-kit/` directory in this fresh worktree — not a code change.)


a-Fig:
Independent review of this PR found two issues, both fixed in bf4298a (on top of the existing 4 commits, pushed to this branch):

**1. Silent group loss in `setGroupSummary` (real bug).** `setGroupSummary` sends `[{ungroup}, {group}]` as one `Truth.apply()` batch so an edit keeps the group's id/handle. `apply()` is not atomic across a batch, though — if the `ungroup` half commits and the paired `group` half is then clamped, the group used to vanish with no error surfaced anywhere.

The most reliable repro isn't the group's own stored range crossing the protect line directly (`pruneProtectedGroups` already guards that case on every housekeeping pass, well before `setGroupSummary` ever runs) — it's `opGroup`'s internal `snappedRange` widening the *edit's* effective range: create a group over an incomplete multi-part assistant message, let a later sibling part of that same message stream in (`pruneProtectedGroups` only checks the group's own already-narrow stored `memberIds`, so this doesn't touch it), move the protect dial past that new sibling but not past the group's own stored range (still safe) — then edit the summary. The regroup re-snaps to the full message, pulls in the newly-protected sibling, and `opGroup` clamps it. The `ungroup` had already committed.

Fixed at the `Truth.apply` layer (covers the live extension path via `t.apply(...)` in `extension/accordion.ts`, not just the app's UI): track any `ungroup` whose derived id (`g:${memberIds[0]}`) a same-batch `group` op targets; if that `group` op is refused, restore the pre-batch group and retroactively flip the `ungroup`'s own `OpResult.applied` to `false`. The `applied` flip matters specifically for live replicas — `appliedOpForWire`/`wireEventFromTruthEvent` in `core/replica.ts` forward only `applied` ops, so leaving the `ungroup` marked applied would forward it alone to a connected replica with no paired revival trigger, silently diverging replica from host (the `rev` counter can't catch this, since both sides still bump it by one). Flipping `applied` makes the whole doomed pair read as "nothing applied," which also correctly skips housekeeping/rev-bump/emit for a batch that net-changed nothing. This is scoped narrowly to the one ungroup-then-matching-derived-id-group pattern — every other multi-op batch keeps its existing (deliberately non-atomic) partial-success semantics, so a conductor batching independent proposals still gets partial success where appropriate.

**2. Missing `.trim()` in `opGroup` (minor, folded into the same fix).** `opGroup`'s human branch stripped fold tags from `op.summary` but never `.trim()`ed the result, unlike its `opFold` sibling. A whitespace-only summary survived as a non-empty, non-null `Group.digest`, so `isDropGroup` didn't recognize it as a drop — but `computeGroupOps` (and `applyPlan`'s own `safeGroups` filter in `core/wire.ts`) both reject a summary that trims to empty, so the group op never reached the wire at all. Net effect: the group still read as folded in the UI while its members shipped whole. Fixed by adding `.trim()` to match `opFold`'s exact pattern — one line.

**Regression tests added** (both confirmed to fail on pre-fix code and pass on post-fix code):
- `app/src/lib/engine/store.digest.test.ts` — reproduces the incomplete-message + append + protect-move + `setGroupSummary` sequence; asserts the group survives with its original `memberIds`/summary rather than being destroyed.
- `core/humanDigest.test.ts` — raw `Truth.apply` with a whitespace-only group summary; asserts `isDropGroup` is true and `computeGroupOps` actually emits a drop op, rather than silently omitting the group.

**Full gate green:**
- `cd app && npm run test` — 844 passed, 20 skipped (pre-existing), 51 files
- `cd app && npx svelte-check --tsconfig ./tsconfig.json` — 0 errors, 0 warnings
- `cd extension && node smoke.mjs` — OK
- `cd extension && node smoke-conductor.mjs` — OK

No existing invariant was touched — all 6 pre-existing `setGroupSummary` tests and the `setBlockDigest` tests in `store.digest.test.ts` pass unmodified, and "the tag is the handle" / asymmetric-empty-semantics behavior is untouched by either fix.

One thing I deliberately did **not** add: a UI-level gate in `Inspector.svelte` mirroring the block editor's `canFoldBlock` check (mentioned as optional defense-in-depth). A naive version (checking the group's *current* stored `memberIds` against `protectedFromIndex`) wouldn't actually catch the real failure mode above, since the risk only appears after `snappedRange` widens the range at edit time — duplicating that widening logic in the UI felt like it would add a false sense of safety without adding real protection, given the Truth-level fix is already airtight. Happy to add it if you'd still like the extra UI signal.

Not merging — leaving that to you.

a-Fig:
Follow-up pass on the adversarial review findings A-D. Pushed 2 new commits on top of `461e0a5` (Finding A, already landed): `7c9b054` and `62bb1c3`.

## Finding A — `groupAgentReachable` over-broad (HIGH)
Already fixed and committed as `461e0a5` before this pass — re-verified unchanged, `Truth.isDegradedDropGroup` distinguishes a genuine role-floor-forced stub from a plain drop, `core/humanDigest.test.ts`'s legitimate role-floor-recap test still passes, and the reviewer's exact repro is covered by a new regression test in `core/store.digest.test.ts`. **Execution-verified.**

## Finding C — `emptied`/cost readout computed from raw draft, not stripped commit value
Confirmed and fixed in `7c9b054`. `emptied` and `draftTokens` were derived from `draft.trim()` — the raw textarea value, which for an engine digest still carries its leading `{#code FOLDED}` tag — instead of `stripFoldTags(draft).trim()`, the exact value `save()` commits. Deleting everything but the tag left the raw draft falsy-non-empty, so the "removes these messages"/"saves as `{empty}`" warning silently never fired even though save() committed a real drop/`{empty}`. Extracted the prediction into `app/src/lib/ui/map/digestEditorLogic.ts` (`committedFromDraft`/`isEmptied`) so `emptied`, `draftTokens`, and `save()` all read off the identical stripped value. 6 new tests in `digestEditorLogic.test.ts`. **Execution-verified**: confirmed failing against the pre-fix `draft.trim()` check, passing against the fix.

## Finding B — `revert()`/Escape doesn't invalidate a stale draft when `typed` was already `null`
Fixed in `7c9b054`, but **verification tier is reasoned-only**, not executed — flagging this explicitly per the task spec since the original finding itself was reasoned-only (reviewer's env was SSR-only).

The mechanism: `revert()` calls `clearDraft(id)` then `typed = null`. If `typed` was *already* `null` on entry (component remounted onto an existing stored draft), that reassignment is a no-op `$state` write — Svelte doesn't notify on an unchanged value — so the `draft` derived kept reading the just-deleted `digestDrafts.ts` Map entry via `getDraft(id)`, leaving the Save/Cancel bar stuck dirty on deleted text. Fix: added a `draftEpoch` version counter (mirrors the `version`/`void this.version` idiom already used in `store.svelte.ts`) that `revert()` bumps and `draft` reads unconditionally, forcing a `getDraft(id)` re-read regardless of whether `typed` itself changed.

Why no execution-verified test: this repo's vitest config runs `environment: "node"` (no jsdom/@testing-library/svelte in devDependencies), so `DigestEditor.svelte` can't be mounted. I made two independent attempts to reproduce the underlying reactivity bug against the compiled `.svelte.ts` module directly and both hit genuine environment limits rather than producing a signal:
1. A harness using bare property-getter reads of a `$derived`/`$derived.by` value (no owning `$effect`) — expected the documented "same-value `$state` write skips notification" behavior to reproduce the staleness. Instead it "self-healed" on both the buggy and fixed code paths identically. Root-caused with a diagnostic call-counter: a bare/non-effect-owned `$derived` read in this Svelte 5.56.1 + vite-plugin-svelte 5.1.1 setup **recomputes unconditionally on every access**, changed or not — a fundamentally different regime from a real mounted component's live-effect dirty-tracking, so the test could never distinguish buggy from fixed.
2. Tried wiring the same harness through a real `$effect.root()` + `$effect()` instead, to get genuine dirty-tracking. Confirmed via `flushSync()` + `await tick()` + `await Promise.resolve()` that **`$effect` callbacks never fire at all** in this environment — no DOM means Svelte's effect scheduler has nothing to run against.

Conclusion: this repo's current Node-only vitest setup cannot genuinely exercise Svelte's live component reactivity/caching semantics either way, which is consistent with (not contradicting) the original reviewer's own caveat. The fix itself is low-risk and matches an established, already-battle-tested codebase idiom, but I want to be explicit that it's verified by code reading, not by a passing/failing test.

## Finding D — deleting a group doesn't clear its stale draft; reused group ids can resurrect it
Confirmed and fixed in `62bb1c3`. Group ids are `g:${memberIds[0]}` and get reused whenever a later group starts at the same leading block. `store.deleteGroup` only ungroups the Truth-side overlay — it never touched `digestDrafts.ts`'s client-local Map, so an unsaved draft could resurrect onto an unrelated later group reusing the same id.

One deliberate deviation from the finding's literal wording ("wherever a group is deleted in the store"): I put the fix in `Inspector.svelte`'s `deleteGroup()` (the sole call site of `store.deleteGroup` in the app — both Delete buttons now route through it) rather than in `store.svelte.ts` itself. `app/src/lib/engine/` has zero existing imports from `app/src/lib/ui/` anywhere in the codebase, and `digestDrafts.ts`'s own doc comment frames itself as UI-only client state deliberately kept out of the engine/Truth layer — `store.svelte.ts` also backs demo/CC/live-replica modes unrelated to the digest-editing UI. Pulling a UI-only draft store into the engine would be a new layering violation for no behavioral difference to the user.

New tests in `inspectorDeleteGroup.test.ts` exercise the real `store.svelte.ts` + `digestDrafts.ts` through a helper mirroring `Inspector`'s `deleteGroup()` exactly, plus a companion pre-fix helper (`store.deleteGroup` alone) that directly reproduces the bug. **Execution-verified** against the real store/drafts modules; the component-level wiring itself (both onclick handlers) is verified by code reading since `Inspector.svelte` can't be mounted in this vitest config either.

## Full gate
- `cd app && npm run test` — 854/854 passing (0 failures)
- `cd app && npx svelte-check --tsconfig ./tsconfig.json` — `0 ERRORS 0 WARNINGS`
- `cd extension && node smoke.mjs` — `smoke: OK`
- `cd extension && node smoke-conductor.mjs` — `smoke-conductor: OK`

Not merging — leaving that to you as usual.

